PR #6, opened 2017-06-23
PAR-101 Updates par_styleguide module page template

nao360:
Displays replica of the GOV.UK Elements index page


kalpaitch:
I don't think it's to do with this, but I'm getting a sass error on compilation. I think @stemount might have already addressed this in which case we should get thoses fixes into master soon.

`[09:28:57] Starting 'sass'...
Error in plugin 'sass'
Message:
    web/themes/custom/par_theme/assets/stylesheets/govuk-template-ie6.scss
Error: It's not clear which file to import for '@import "govuk-template"'.
       Candidates:
         govuk-template.scss
         govuk-template.css
       Please delete or rename all but one of these files.
        on line 5 of web/themes/custom/par_theme/assets/stylesheets/govuk-template-ie6.scss
>> @import "govuk-template";`

kalpaitch:
Ah ok, so he did remove them, but then running gulp recreates the .css files. They're not committed to the repo but running gulp again will still fail because they exist locally.

```

**Comment by kalpaitch** [comment]
```

kalpaitch:
So as you can see from my last commit, I restructured the stylesheet directories a bit to separate sass and css.

This is mostly done according with the suggestions in the govuk-elements-sass npm package. And without much effort the page looks great:

image

gulpfile.js
```diff
@@ -5,19 +5,24 @@ var gulp = require('gulp'),
...
- return gulp.src(themeDir + '*.scss')
-  .pipe(sourcemaps.init())
-  .pipe(sass().on('error', sass.logError))
-  .pipe(sourcemaps.write(themeDir))
-  .pipe(gulp.dest(themeDir));
+  return gulp.src(themeDir + '/sass/**/*.scss')
+    .pipe(sourcemaps.init())
+    .pipe(sass({
+      includePaths: [
+        'node_modules/govuk_frontend_toolkit/stylesheets', // 1
```
kalpaitch: Just wanted to point out that we're importing alphagov styles from the npm packages now.

PR #8, opened 2017-06-29
PAR-130: Added file handling to multi-step forms.

kalpaitch:


nao360:
@kalpaitch 

Re: orphaned file removal - is '1 week' based on a known requirement (eg., NFR), or is it a simply a starting value, and to be confirmed in due course? Thanks.

nao360:
@kalpaitch are we ignoring travis CI build fails for the time being? If so, I'm happy to approve.

PR #16, opened 2017-07-03
Par 137 - Vagrant development environment

chris-moreton:
@stemount @jaffamonkey Ready to test


stemount:
@chris-moreton do you think the ubuntu user should be added to the Docker group so sudo commands are not needed? ta 

image

https://docs.docker.com/engine/installation/linux/linux-postinstall/

stemount:
Chris - I have decided that this should be in master with urgency, the good news is this works on my machine nicely. 👍 💯

PR #20, opened 2017-07-04
Cucumber

jaffamonkey:
Updated features


jaffamonkey:
The new html report (with screenshot) is generated in: ./tests/reports/html/index.html

PR #29, opened 2017-07-10
PAR-103

stemount:
Styling for pagerer module. Please use https://rural-payments-styleguide.herokuapp.com/pagination/ as a reference guide on how to check this. Please check mobile viewports.

A gulp build will be required (or just use the rebuild script)


kalpaitch:
image

PR #38, opened 2017-07-12
@naoyoshino improving as per your JIRA comment - yes, these should ha…

stemount:
…ve styles that match the header.


stemount:
@nao360 maybe this solution is better, extending the styles rather than copy and pasting? this way we can ensure the GDS a proposition-links' changes can be inherited.

PR #49, opened 2017-07-13
PAR-36 Responsive scrolling tables on mobile/tablet

stemount:
See GIF

par-36


nao360:
@stemount I think as a separate story we might look at locking the first column (ie., Primary Authority) so that in mobile view, when scrolling the remaining columns to the left there is some reference as to what PA the row belongs to. I know, I had to re-read that to make sure it made sense. It does! Good work, buddy, really like the way tables are looking in mobile view.

kalpaitch:
In tables generated by views they have this priority thing going on so that you can hide columns that aren't important, that might also be useful for some...

Also just to add that in doing this ticket @stemount realised we had no documentation of what our browser requirements are :)

stemount:
@kalpaitch aye, as per documentation https://drupalize.me/blog/201409/how-add-responsive-tables-content-drupal-8, just testing this _actually_ works.

PR #52, opened 2017-07-14
Use the govuk_template "HTML version" assets in a custom `images/PAR`…

stemount:
… folder because the ejs/mustache versions on npm registry do not contain these assets so they cannot be copied/inherited in the assets folder using a Gulp copy mechanism.


stemount:
image

for testing: it should look like the above.

PR #55, opened 2017-07-14
form password fields should also have the form-control attribute.

stemount:
log_in___regulatory_authority

improve the login form by simply applying a CSS class


kalpaitch:
Just so you're aware there will likely be other such fields that we haven't caught yet. Nice work.

PR #56, opened 2017-07-17
PAR-205: Cleanse - removing old par entities.

kalpaitch:
This is required before we can add the new ones.


stemount:
Makes sense, checking passes against master, but assume it would be best to get this into master ASAP.

kalpaitch:
Exactly @stemount

PR #57, opened 2017-07-17
update primary/secondary links template for correct classes.

stemount:
1. override the Primary/Secondary links template so the Primary/Secondary menu links have the `links` class. e.g. `<ul class="links">`

https://govuk-elements.herokuapp.com/typography/#typography-inset-text

This ticket will be fully closed when we refactor the GDS SASS inheritance, as the ul, li {} _should have no bullets by default_, alas they don't.

For now, this is what is expected from this PR.

log_in___primary_authority_register


stemount:
@nao360 GDS style inheritance refactor is not in Sprint 2 scope right now, will discuss with you in a sec.

PR #58, opened 2017-07-17
Patch Drush for compatibility for pg_dump w/ Postgres 9.5

stemount:
Implementing a patch from https://github.com/drush-ops/drush/pull/2282/commits/145dd25e5d6b037e59c0a457d42688a2ccc74ee0 to fix compatibility with Postgres 9.5+.

Updating composer.* files for later versions.


stemount:
as per our IRL discussion @kalpaitch… decided having `./patches` folder for things that are out of control e.g. the PR patch could be littered with other code we don't want as the PR is still open.

PR #63, opened 2017-07-18
Dump sanitized and unsanitized versions of database:

chris-moreton:
Phase one - perform the dumps.


web/sites/all/drush/par.aliases.drushrc.php
```diff
@@ -334,11 +334,23 @@
...
   'root' => '/home/vcap/app/web'
 );
-$aliases['prod'] = array(
-  'user' => 1,
-  'root' => '/home/vcap/app/web'
-);
\ No newline at end of file
+$aliases['paas-unsanitized'] = array(
+    'user' => 1,
+    'command-specific' => array (
```
kalpaitch: Probably worth removing. Otherwise all good.

PR #64, opened 2017-07-18
PAR-172: Added stub entities.

kalpaitch:
To enable stub entities, install the par_data_test module. This is enough to get test content into the site.

NOTE: Uninstalling this module deletes all par data and refreshes to original state.

If you want only stub entities to be used. set the stub value to true in par_data.settings.yml and choose which entity to load for each type in the same file.


kalpaitch:
Requires https://github.com/TransformCore/beis-par-beta/pull/65

PR #65, opened 2017-07-18
PAR-245: Added all required relationships.

kalpaitch:


stemount:
is it really good enough for me to just review these PRs with just a visual check? we don't build branches right now so its a bit of an approval lottery when grand changes are made like this…

trashing local machines maybe is perhaps a bit too slow to do this…

kalpaitch:
Good spots @stemount

stemount:
Just earning my spots/stripes @kalpaitch

PR #70, opened 2017-07-20
Drush commands to upload and download from S3

chris-moreton:


patches/flysystem-drush-command.patch
```diff
@@ -0,0 +1,87 @@
...
++      ],
++      'examples' => [
++          'drush fsp s3public tests/report.json' => 'Put tests/report.json on the s3public Flysystem',
++      ],
++  ];
++  $commands['flysystem-get'] = [
++    'description' => 'Get a file from a Flysystem stream',
++    'aliases' => ['fsg'],
++    'arguments' => [
++      'arguments' => [
```
kalpaitch: Don't need the second 'arguments' array.

patches/flysystem-drush-command.patch
```diff
@@ -0,0 +1,87 @@
...
++/**
++ * Implements hook_drush_command().
++ */
++function flysystem_drush_command() {
++  $commands = [];
++  $commands['flysystem-put'] = [
++    'description' => 'Upload a file to a Flysystem stream',
++      'aliases' => ['fsp'],
++      'arguments' => [
++          'filepath' => 'Path to the file on the local file system, including filename',
```
kalpaitch: 4 space indentation instead of 2 :)

patches/flysystem-drush-command.patch
```diff
@@ -0,0 +1,87 @@
...
++    
++    $flysystem = get_fs_scheme($scheme);
++    $resource = fopen($localPath, 'r');
++    $flysystem->writeStream($remotePath, $resource);
++    fclose($resource);
++    
++    return $localPath . ' written to ' . $scheme . ':' . $remotePath;
++}
++
++function drush_flysystem_get($scheme = null, $remotePath = null, $localPath = null)
```
kalpaitch: Also just putting for reference... http://docs.aws.amazon.com/aws-sdk-php/v2/guide/feature-s3-stream-wrapper.html#downloading-data

It might be good to start using `fopen` and iterate over a given number of bytes as you mentioned. Same for put.

PR #72, opened 2017-07-20
Cucumber

jaffamonkey:
Added very basic check of errors (by count). In ref to PAR-203/PAR-173.


tests/src/features/PAR189-Edit-Main-PA-Contact-Save-Form-Data.feature
```diff
@@ -0,0 +1,22 @@
+@Pending
```
kalpaitch: @jaffamonkey sorry I had to pause this test, it's better to get the rest of your tests in. It will fail because the form is cached for non-logged in users. If you want to test any of the PAR Journeys you have to be logged in as someone.

jaffamonkey:
No worries - maybe too quick to add that to CI just yet!

PR #74, opened 2017-07-21
Styleguide Components for PAR-9 /styleguide/components

stemount:
Please leave this open for the time being, this will grow with time.

@jaffamonkey /styleguide/components is a new page added to .pa11yci as well


stemount:
we may have a leaner way to do with this Entity view modes e.g. teaser…

/styleguide/components
image

web/modules/custom/par_styleguide/par_styleguide.info.yml
```diff
@@ -3,3 +3,5 @@ type: module
 description: Describes all the components and their styles according to the GOV.UK style guidelines
 core: 8.x
 package: PAR
+dependencies:
```
kalpaitch: Cool, if we're doing this I think we need to add the par_components to the core.extension.yml file. All looks good otherwise.

kalpaitch:
I think this is super cool. We may be able to just use view modes to get the fields so that we don't have to get each field separately, but that's super groundwork.

web/modules/custom/par_styleguide/par_styleguide.info.yml
stemount: agreed

PR #76, opened 2017-07-21
PAR-179: Re-oganised Flows and added interfaces and readme.

kalpaitch:


web/modules/custom/par_flows/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+As can be seen forms should specify their form_id, which will be used as part of the key to save and retrieve any data they wish to set.
+
+The data they set is stored in the Private Temp Store and can be retrieved only by another Form in this Flow. Pages are not able to access this data.
+
+## Base Classes
+The routes defined in a Flow must implement either the ParBaseForm or ParBaseController abstract classes, or they will not be able to access the transitions or store/retrieve data.
+
+* Data storage/retrieval methods are proteted and as such only documented in the ParBaseForm method.
```
stemount: proteted functions are my favourite

PR #77, opened 2017-07-21
PAR-9

stemount:
List of Tasks for a Primary Authority
Route: /dv/primary-authority-partnerships/%authority/partnership/%partnership

Early PR to check implementation makes sense

image


stemount:
@nao360 this page is actually in the build

*Enable stub entities*
`docker exec -it par_beta_web /var/www/html/vendor/bin/drush --root=/var/www/html/web en -y par_data_test`
`docker exec -it par_beta_web /var/www/html/vendor/bin/drush --root=/var/www/html/web config-set par_data.settings stubbed true`

then visit 
http://localhost:8111/dv/primary-authority-partnerships/1/partnership/1/details

web/modules/custom/par_data/src/Entity/ParDataPerson.php
```diff
@@ -156,6 +156,23 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
       ])
       ->setDisplayConfigurable('form', FALSE);
 
+    // Role.
```
kalpaitch: Hi, yea role isn't in the physical model yet. I like this, but I think we'll need to let Jamie know that the UX wants to capture this extra information before we put it in. Remove for now.

Also we'll do all updates to the data model in one go, and we have a few that Jamie is working on. Nice spot though.

web/modules/custom/par_data_test/par_data_test.module
```diff
@@ -148,7 +148,8 @@ function par_data_test_create_entities() {
     'person_name' => 'Big Bird',
     'work_phone' => '01723456789',
     'mobile_phone' => '0777777777',
-    'email' => 'bigbird@example.com'
+    'email' => 'bigbird@example.com',
+    'role' => 'CTO'
```
kalpaitch: Same again. Let's wait on this.

web/modules/custom/par_flow_transition_partnership_details/src/Controller/ParFlowTransitionPAListOfTasks.php
```diff
@@ -0,0 +1,181 @@
...
+use Drupal\par_data\Entity\ParDataAuthority;
+use Drupal\par_data\Entity\ParDataPartnership;
+use Drupal\par_data\Entity\ParDataOrganisation;
+use Drupal\par_flows\Form\ParBaseForm;
+use Drupal\par_flows\ParBaseInterface;
+
+/**
+ * A controller for all PAR Flow Transition pages.
+ */
+class ParFlowTransitionPAListOfTasks extends ParBaseForm {
```
kalpaitch: Cool, like you said, if you extend ParDataController instead of ParBaseForm that would be ace.

web/modules/custom/par_flow_transition_partnership_details/src/Controller/ParFlowTransitionPAListOfTasks.php
```diff
@@ -0,0 +1,181 @@
...
+   */
+  protected $flow = 'transition_partnership_details';
+
+  /**
+   * Returns a unique string identifying the form.
+   *
+   * @return string
+   *   The unique string identifying the form.
+   */
+  public function getFormId() {
```
kalpaitch: Don't need this method when extending ParDataController I think.

web/modules/custom/par_flow_transition_partnership_details/src/Controller/ParFlowTransitionPAListOfTasks.php
```diff
@@ -0,0 +1,181 @@
...
+    $this->loadDataValue("person_{$primary_person->id()}_role", $primary_person->get('role')->getString());
+
+  }
+
+  /**
+   * @param ParDataPartnership $partnership_status
+   *
+   * @return boolean
+   */
+  public function arePartnershipDetailsConfirmed(ParDataPartnership $partnership) {
```
kalpaitch: You're dead right, I feel this method is kinda generic to all ParData entities so was gonna pull back and put on the entity class itself. I'll do that shortly.

web/modules/custom/par_flow_transition_partnership_details/src/Controller/ParFlowTransitionPAListOfTasks.php
```diff
@@ -0,0 +1,181 @@
...
+    $partnership_status = $partnership->get('partnership_status')->getString();
+
+    return ($partnership_status == 'Approved') ? true : false;
+
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state, ParDataAuthority $par_data_authority = NULL, ParDataPartnership $par_data_partnership = NULL) {
```
kalpaitch: When we extend ParDataController this method will most likely become 'content' and will not have the two form parameters on it.

web/modules/custom/par_flow_transition_partnership_details/src/Controller/ParFlowTransitionPAListOfTasks.php
```diff
@@ -0,0 +1,181 @@
...
+  /**
+   * Helper to get all the editable values when editing or
+   * revisiting a previously edited page.
+   *
+   * @param ParDataAuthority $par_data_authority
+   *   The Authority being retrieved.
+   * @param ParDataPartnership $par_data_partnership
+   *   The Partnership being retrieved.
+   */
+  public function retrieveEditableValues(ParDataAuthority $par_data_authority = NULL, ParDataPartnership $par_data_partnership = NULL) {
```
kalpaitch: And because this method also uses loadDataValue which is a method of ParBaseForm only, so we might need to do this slightly differently. Let's just chat about this though. Ace Work.

kalpaitch:
@jaffamonkey These tests are failing because of access permissions. Can we get all PAR journey tests running as the logged in user:
**testuser**: **testpwd**

PR #80, opened 2017-07-24
Par 240

stephenb01:


kalpaitch:
Looks good, it does display a gray background to the WYSIWYG though, anywhere there isn't text basically:
screenshot from 2017-07-24 15-55-14

PR #82, opened 2017-07-24
remove whitespace & drush sql-drop before sql-cli /sync

stemount:
I often run into problems with xmlsitemap and watchdog PDOexceptions saying table already exists in database without having a clean postgres database.

image


kalpaitch:
Better get confirmation from the Mothership before we merge this :+1:

PR #97, opened 2017-07-26
Dev: Locking down application headers for security

kalpaitch:
@nao and I had a conversation earlier about iFraming content, which made me realise we need to lock down our security headers - https://securityheaders.io/

Enabling a very simple module (one wot I wrote earlier) to adapt some potentially leaky headers.


sync/http_response_headers.response_header.x_powered_by.yml
```diff
@@ -0,0 +1,12 @@
+uuid: b6c5660e-ab6f-4caf-9213-c10c770cdd32
+langcode: en
+status: true
+dependencies: {  }
+_core:
+  default_config_hash: _Bau76mGw4kCqnHi4BTCC-Uq2PS0Ih5hZ94lblLewEE
+id: x_powered_by
+label: X-Powered-By
+description: 'The X-Powered-By header gives information on the technology that''s supporting the Web Server. It is best not to provide this information.'
```
stemount: does this escape the `'`?

Nao:
@kalpaitch, I don't think that's me you were referring to..?

kalpaitch:
@Nao It was vaguely related to iFraming content for the marketing site. Not a biggie, but not an issue no more.

Nao:
Yes, but I have no idea what 'TransformCore' is, or what 'iFraming' is, so I doubt it's me really.. ;)
I'm the author of Wedge, Kyodai and Lestrade's. If you don't know those either, chances are you poked the wrong person. ;)

PR #99, opened 2017-07-27
Dev: Notify integration

kalpaitch:
@nao360 on the off chance that you do want notify integration, this is something I kicked up whilst watching Dirty Grandpa!! (aka. Sean Joyce)

I realised we had no capabilities to send any system emails from the site and figured this could be quite useful.

image


kalpaitch:
@nao360 Hopefully not but let's see once he's got his current S3 restore ticket in...

PR #106, opened 2017-07-28
PAR-48: User roles and security settings.

kalpaitch:
Enabled the register with e-mail options so that now instead of a username box on the login & registration forms it's now a username or e-mail box.

Enabled the password policy module which adds some restrictions for administrator passwords, but more importantly expires all passwords (administrators after 6 months, other PAR users after 1 year).


composer.json
```diff
@@ -19,17 +19,19 @@
...
         "drupal/csv_serialization": "~1.0.0",
+        "drupal/email_registration": "~1.0.0",
         "drupal/features": "~3.5.0",
         "drupal/file_entity": "~2.0.0@beta",
         "drupal/flysystem": "~1.0.0",
         "drupal/flysystem_s3": "~1.0.0@alpha",
         "drupal/govuk_notify": "1.x-dev",
         "drupal/health_check": "1.0.0",
-        "drupal/http_response_headers": "dev-2.x",
+        "drupal/http_response_headers": "~2.0.0",
```
stemount: woooo :)

stemount:
@kalpaitch is this ready to go?

kalpaitch:
Re-enabling the par_data_test module now and the correct permissions and roles are assigned. And the users are funnelled down their correct paths...

image

If there happens to be an edge case where their email address is used for multiple roles this is the order of precedence: Authority, Business, Coordinator.

PR #120, opened 2017-07-30
PAR-129: Invite links sent out.

kalpaitch:
@nao360 It turns out it was easier to just send the e-mail using a token than having to get the invite service, create a new link and then display it for them to copy paste :) We can definitely add that in so they can copy the message body and send it how they wish, but will get everything else working first.

screenshot from 2017-07-30 20-49-47

image

screenshot from 2017-07-30 21-09-10

Woop woop!!!


kalpaitch:
Happy with this PR as is at the moment, the sending of an invite works well.

There's still a little bit of work to do on the 'accepting' of an invite, the module isn't as complete as I would like it to be, but it's a lot better than building one ourselves. At the moment when you click to accept an invite you go to the registration page. The invite process doesn't change the default registration process, so either visitors get an access denied on the registration form if only admins can create an account. Or they have to verify their account by email again (they've just clicked a link from their email), or they have to wait for administrator approval again. So we need to improve on this.
screenshot from 2017-07-30 21-59-03
Also we need to make it so that they can't change the email address when registering.

kalpaitch:
See https://regulatorydelivery.atlassian.net/browse/PAR-279 for ticket for further work.

PR #121, opened 2017-07-31
PAR-75: Added edit screen for trading name in business journey.

stephenb01:


web/modules/custom/par_flow_transition_business/src/Form/ParFlowTransitionTradingForm.php
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state, ParDataPartnership $par_data_partnership = NULL, $trading_name_delta = NULL) {
+    $this->retrieveEditableValues($par_data_partnership);
+    $par_data_organisation = current($par_data_partnership->get('organisation')->referencedEntities());
```
kalpaitch: Just realised looking at this that we're relying on there being only one $par_data_organisation on a partnership. I guess really we should have put the organisation id in the url somewhere to allow us to be more flexible. Not important for this, it's only transition, but one to think about for beta.

stephenb01: Agreed.

PR #125, opened 2017-07-31
PAR-13: Added 6 test files, for use in advice and inspection plans

kalpaitch:
There is not yet a 'Document' field on either Par Advice or Par Inspection Plans so these documents don't get attached to anything at the moment.

@stephenb01 When the physical model is updated can you add a few of these files to the par_data_advice  test entity and one to the inspection plan entity. Thanks!! :)

If you want to go directly to the Document Meta Data Form go to this link directly for the time being - `/dv/primary-authority-partnerships/1/document/1`


web/modules/custom/par_data_test/par_data_test.install
```diff
@@ -28,6 +28,20 @@ function par_data_test_uninstall() {
...
+      ->getModule('par_data_test')->getPath() . '/assets/';
+  $files = scandir($path);
+  if (!$files) {
+    $files = [];
+  }
+  foreach (\Drupal::entityTypeManager()->getStorage('file')->loadMultiple() as $file) {
+    if (in_array($file->getFilename(), $files)) {
+      $file->delete();
+    }
+
```
stephenb01: not 100% certain this will work as expected. Looks like you get the file names from the assets directory and then compare the name names taken from the file storage and if they match then remove it. What happens if I have added a file in a different directory of the same name?

web/modules/custom/par_flow_transition_partnership_details/src/Form/ParFlowTransitionAdviceForm.php
```diff
@@ -36,6 +36,20 @@ public function retrieveEditableValues(ParDataPartnership $par_data_partnership
...
+      $allowed_types = $par_data_advice->type->entity->getConfigurationByType('advice_type', 'allowed_values');
+      // Set the on and off values so we don't have to do that again.
+      $this->loadDataValue('allowed_types', $allowed_types);
+      $advice_type = $par_data_advice->get('advice_type')->getString();
+      if (is_array($allowed_types) && in_array($advice_type, $allowed_types)) {
+        $this->loadDataValue('document_type', $advice_type);
+      }
+
+      // @TODO We need to work out how to get a list of regulatory functions.
+      if ($we_know_the_regulatory_functions = FALSE) {
```
stephenb01: Shouldn't this be:
if ($we_know_the_regulatory_functions == FALSE) {

web/modules/custom/par_data_test/par_data_test.install
stemount: I agree, this is a test module though that will never see production I guess.

kalpaitch: The `getFilename()` returns just the name not the path apparently. It seems to remove them.

web/modules/custom/par_flow_transition_partnership_details/src/Form/ParFlowTransitionAdviceForm.php
kalpaitch: Well yes, but we don't know how to relate them to advice documents, that was the big question mark on the Physical model document we're looking at... We'll have to leave this one like this until we can get some feedback from Jamie or maybe Ben.

stemount: this is in @todo block as well, it's clearer to me a placeholder, so ignoring.

PR #127, opened 2017-07-31
PAR-195: Added Inspection Plan Form

kalpaitch:


web/modules/custom/par_flow_transition_partnership_details/src/Form/ParFlowTransitionAdviceForm.php
```diff
@@ -8,10 +8,9 @@
 use Drupal\par_flows\Form\ParBaseForm;
 
 /**
- * The primary contact form for the partnership details steps of the
- * 1st Data Validation/Transition User Journey.
+ * The advice document form for Transition Journey 1.
  */
-class ParFlowTransitionAdviceForm extends ParBaseForm {
+class ParFlowTransitionInspectionPlanForm extends ParBaseForm {
```
stemount: yo @kalpaitch,

re: `Fatal error: Cannot declare class Drupal\par_flow_transition_partnership_details\Form\ParFlowTransitionInspectionPlanForm, because the name is already in use in /var/www/html/web/modules/custom/par_flow_transition_partnership_details/src/Form/ParFlowTransitionAdviceForm.php on line 128`

is this supposed to be `ParFlowTransitionAdviceForm`?

PR #129, opened 2017-08-01
Cucumber

jaffamonkey:
Important test updates (i.e. some tests were failing incorrectly)


tests/wdio.BROWSERSTACK.conf.js
```diff
@@ -6,14 +6,18 @@ config.capabilities = [{
...
+        debug: true
 }];
 
 config.services = ['browserstack'];
 config.user = 'euniceaidoo1';
 config.key = 'z8aabpqRzKvWZDHKTKff';
 config.browserstackLocal = true;
 config.host = 'hub-cloud.browserstack.com';
-config.baseUrl = 'https://localhost:80';
+config.baseUrl = 'https://par-beta-test.cloudapps.digital';
```
kalpaitch: @jaffamonkey Is this intentional? Pointing it at our live site?

jaffamonkey: That is intentional - more pragmatic as tests commonly fail testing localhost urls, down to severe lag time.

jaffamonkey: This only affects cucumber js run against browserstack service

tests/wdio.BROWSERSTACK.conf.js
```diff
@@ -6,14 +6,18 @@ config.capabilities = [{
...
         os: 'Windows',
         os_version: '7',
-        resolution: '1024x768'
+        resolution: '1024x768',
+        debug: true
 }];
 
 config.services = ['browserstack'];
 config.user = 'euniceaidoo1';
 config.key = 'z8aabpqRzKvWZDHKTKff';
```
kalpaitch: Also, is this a live access key?? If so we'll need to remove from public repo and revoke key??

PR #131, opened 2017-08-01
Inspired by Bootstrap, `btn-class` applied to <button /> will make bu…

stemount:
…ttons to look/act like links.

use cases:
journey1 step 1 view exposed filters.
for pages which have more than one button.


stemount:
FYI its this Update button on the left sidebar @nao360 
image

PR #140, opened 2017-08-03
Advice list ticks helptext and headers

stemount:
Add a HTML tick instead of 100% completion percentage.

image


kalpaitch:
Nice work @stemount Just fixed a few access issues and added the document download link...

image

PR #141, opened 2017-08-03
PAR-281: DV Journey 1 Step 1 Filter/Sort

stemount:
I think this is perhaps blocked by the views relationship thing.


kalpaitch:
Looks good, but I'll let someone else test this on their local please @nao360 @stephenb01 because I can't afford to switch of the migration branch at the moment.

PR #143, opened 2017-08-03
coding standards and making links uniform using #prefix. This <br> bu…

stemount:
…siness will end up being <p> tags where there is time to finesse the theme in future sprints.


web/modules/custom/par_flow_transition_partnership_details/src/Form/ParFlowTransitionTermsForm.php
```diff
@@ -56,7 +58,7 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
       '#markup' => "<p>Please Review the new Primary Authority terms and conditions and confirm that you agree with them.</p><p>The New terms will come into effect from <em>01 October 2017</em>.</p><p>What's changed?</p>",
     ];
 
-    // @stemont will need your input on componentizing this summary box.
+    // @todo stemont will need your input on componentizing this summary box.
```
kalpaitch: Fair enough, Happy to chat, I guess I didn't really know what to do with it either though :)

PR #146, opened 2017-08-07
PAR-239: Updated entities and methods for accessing configuration.

stephenb01:


web/modules/custom/par_data/src/Entity/ParDataLegalEntity.php
```diff
@@ -64,6 +64,14 @@
  */
 class ParDataLegalEntity extends ParDataEntity {
 
+  public function getLegalEntityTypes() {
```
kalpaitch: Would be good to decide how we really want to define configuration and validation rules...

a) We currently have config on the entity types - https://github.com/TransformCore/beis-par-beta/blob/master/sync/par_data.par_data_partnership_type.partnership.yml#L17
b) We could use select field storage and put it on that.

I guess I'm reluctant to put it in code... :)

web/modules/custom/par_data/src/Entity/ParDataPerson.php
```diff
@@ -224,25 +224,63 @@ public function isAuthorityMember() {
...
+  /**
+   * Get list of preferred communication methods.
+   * @return array
+   */
   public function getPreferredCommunicationMethods() {
     return [
-      'Work phone',
-      'Mobile Phone',
-      'Email',
+      'phone' => 'Work phone',
```
kalpaitch: Also with this we can have a 'communication_preference_fields' on the entity type configuration property, much the same way as the 'required_fields' property that is already there... https://github.com/TransformCore/beis-par-beta/blob/master/sync/par_data.par_data_partnership_type.partnership.yml#L14

web/modules/custom/par_flow_transition_business/src/Form/ParFlowTransitionLegalEntityForm.php
```diff
@@ -79,18 +79,12 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
...
-      'Mr',
-      'Dr',
-    ];
+    $options = $par_data_legal_entity->getLegalEntityTypes();
     $form['Legal_entity_type'] = [
       '#type' => 'select',
       '#title' => $this->t('Type of Legal Entity'),
       '#default_value' => isset($id) ? $this->getDefaultValues("legal_entity_{$id}_legal_entity_type") : '',
-      '#options' => array_combine($title_options, $title_options),
+      '#options' => array_combine($options, $options),
```
kalpaitch: I don't think we want to array_combine for options. We want to save the value as the key, but display the value as the value. Probably worth putting a method on ParDataEntity to do this??

sync/par_data.par_data_legal_entity_type.legal_entity.yml
```diff
@@ -8,4 +8,9 @@ id: legal_entity
 label: 'Legal Entity'
 description: 'A Legal Entity is a representation of an Organisation via some formal method of registration or else a less formal declaration.'
 help: ''
-configuration: null
+configuration:
+  legal_entity_types:
+    values:
```
kalpaitch: The key should really be `allowed_values`. `values` isn't currently supported by the entity schema...

```
# Allowed configuration values are listed below.
par_data.configuration_type.status_field:
```
which identifies which is the main status field for use when we do `$par_entity->getParStatus()` OR
```
par_data.configuration_type.required_fields:
```
which currently determines which fields are required and which is used to identify how complete an entity is using $par_entity->getCompletionPercentage() OR
```
par_data.configuration_type.allowed_values:
```
which is used to get the allowed values... I think this is what you want.

sync/par_data.par_data_person_type.person.yml
```diff
@@ -8,4 +8,15 @@ id: person
...
 help: ''
-configuration: null
+configuration:
+  partnership_preferred_communications:
+    values:
+      phone: 'Work Phone'
+      mobile: 'Mobile Phone'
+      email: 'Email'
+  title_options:
+    values:
```
kalpaitch: `allowed_values`

sync/par_data.par_data_person_type.person.yml
```diff
@@ -8,4 +8,15 @@ id: person
 label: Person
 description: 'A Person is a named individual who can feature in a number of different ways within PAR.  A Person may, or may not, be a user of the PAR application.'
 help: ''
-configuration: null
+configuration:
+  partnership_preferred_communications:
+    values:
```
kalpaitch: `allowed_values`

sync/par_data.par_data_premises_type.premises.yml
```diff
@@ -8,4 +8,10 @@ id: premises
 label: Premises
 description: 'Premises are a location used by either an Authority or an Organisation.'
 help: ''
-configuration: null
+configuration:
+  country_options:
+    values:
```
kalpaitch: `allowed_values`

web/modules/custom/par_data/src/Entity/ParDataPerson.php
```diff
@@ -224,25 +224,69 @@ public function isAuthorityMember() {
...
+   * Get list of preferred communication methods.
    *
-   * @return mixed|null
-   *   Returns any business records found.
+   * @return array
    */
-  public function setPreferredCommunication($method) {
-    $methods = ['phone', 'mobile', 'email'];
+  public function getPreferredCommunicationMethods() {
+    return $this->type->entity->getConfigurationByType('partnership_preferred_communications', 'allowed_values');
```
kalpaitch: This seems like a weird one, may have a quick chat tomorrow. I guess we're treating it as one form element with multiple fields. Not sure 'allowed_values' really fits the case for boolean fields.

stephenb01: This one confused me as well. Not 100% certain what to do in this case.

PR #147, opened 2017-08-07
PAR-281: Add Status filter to DV Journey 1 Step 1

stemount:
Keeping this as lean as possible.
Enabling Better Exposed Filters as I know this will be used for everything else to keep it uniform.

image


stemount:
@kalpaitch this was already snook in earlier, 
image

PR #148, opened 2017-08-07
fixing regression on PAR-332.

stemount:
before 
image

after

image


stemount:
PR #151 will sort this out

PR #153, opened 2017-08-08
Update authority to use ParDataViewsData, which has also been updated.

stemount:
as needed for RD Helpdesk and Partnership view…


stemount:
@kalpaitch not ever, or not now? 😀

kalpaitch:
Oh no defo want this in, but that class contains some not ready for production code. Feel free to work on getting the additional views data fields in there.

PR #156, opened 2017-08-08
PAR-40: RD Helpdesk WIP

stemount:
This is WIP. It is functional, but needs to be tested with a bigger dataset to test pagination and also tweak the amount of items per page.


stemount:
image

stemount:
image

web/modules/custom/par_data/src/Entity/ParDataAuthority.php
```diff
@@ -26,7 +26,7 @@
  *     "storage_schema" = "Drupal\trance\TranceStorageSchema",
  *     "view_builder" = "Drupal\Core\Entity\EntityViewBuilder",
  *     "list_builder" = "Drupal\trance\TranceListBuilder",
- *     "views_data" = "Drupal\par_data\Views\TranceViewsData",
+ *     "views_data" = "Drupal\par_data\Views\ParDataViewsData",
```
kalpaitch: @stemount Just need to update that class a bit, will have a look now.

web/modules/custom/par_flow_transition_partnership_details/par_flow_transition_partnership_details.module
```diff
@@ -5,18 +5,85 @@
...
   }
 
 }
+
+/**
+ * Implements hook_views_pre_render().
+ *
+ * @param \Drupal\views\ViewExecutable $view
+ */
+function par_flow_transition_partnership_details_views_pre_render(ViewExecutable $view) {
```
kalpaitch: Can we create a new module called par_dashboards for all this stuff. It's got very little to do with this Primary Authority User Journey. Sorry for being fussy, but would be really good to keep everything separate.

web/modules/custom/par_flow_transition_partnership_details/par_flow_transition_partnership_details.module
```diff
@@ -5,18 +5,85 @@
...
+function par_flow_transition_partnership_details_views_query_alter(ViewExecutable $view, $query) {
+
+  // PostgreSQL case insensitive query alter.
+  foreach ($query->where as $group_key => $group) {
+    foreach ($group['conditions'] as $key => $condition) {
+
+      // Only apply to combined fields, Views stamps in queries are useful!
+      if (preg_match('/:views_combine/', $condition['field'])) {
+        // ILIKE is proprietary to PostgreSQL.
+        $query->where[$group_key]['conditions'][$key]['field'] = str_replace('LIKE', 'ILIKE', $condition['field']);
```
kalpaitch: Tying ourselves to a particular backing service is very much against the principles, not convinced we should do this??

stemount: I agree, will keep in the history, but it is an enhancement, so if this comes back, it could work out if the active database is Postgres.

stemount:
@kalpaitch agreed, split into par_dashboards.

PR #157, opened 2017-08-09
Update test users to also include another multi-role user.

stemount:


stemount:
Meh. This just fails.

PR #168, opened 2017-08-11
Cucumber

jaffamonkey:
Added some coordinator journeys, and a few other checks - this work absolutely not done, just to exceed the current unit test count. The fact it does is pure coincidence


.travis.yml
```diff
@@ -65,6 +65,8 @@ install:
   - echo "PAR_HASH_SALT=$PAR_HASH_SALT" >> .env
   - echo "DATAFILE=$DATAFILE" >> .env
   - echo "APP_ENV=travis" >> .env
+  - echo "BSK_USER=euniceaidoo1" >> .env
+  - echo "BSK_KEY=z8aabpqRzKvWZDHKTKff" >> .env
```
kalpaitch: Just need to get these variables encrypted into Travis

PR #172, opened 2017-08-12
as settings.local.php is not working anymore, how about this to preve…

stemount:
…nt local caching - i have to put this file back every time - just an idea


kalpaitch:
I think better to get composer to copy the `example.settings.local.php` on post-install.

PR #173, opened 2017-08-12
Create dynamic full name field for "Summary" view of a PAR Person entity.

stemount:
PSR-4, create dynamic field.


web/modules/custom/par_data/par_data.module
```diff
@@ -6,37 +6,72 @@
...
+}
+
+/**
+ * Implements hook_entity_view_alter().
+ *
+ * @param $build
+ * @param Entity            $entity
+ * @param EntityViewDisplay $display
+ */
+function par_data_entity_view_alter(&$build, Entity $entity, EntityViewDisplay $display)
```
kalpaitch: We shouldn't have to use hooks for this kinda stuff. We have the ViewsData class for this.

Also I'm particularly conscious that we don't want to 'code in' any business logic, like combining name fields. This should all be configurable. Got to work quite hard to keep the business logic separate from the code but it will get messy quickly otherwise. It's not too much of an issue for anything we're only using for transition.

kalpaitch:
Also, instead of using views fields I think it would be more useful to create new pseudo fields and/or field formatters that we could use anywhere.

web/modules/custom/par_data/par_data.module
```diff
@@ -6,37 +6,72 @@
...
+
+use Drupal\par_data\Entity\ParDataPerson;
 
 /**
  * Implements hook_help().
  */
-function par_data_help($route_name, RouteMatchInterface $route_match) {
-  switch ($route_name) {
+function par_data_help($route_name, RouteMatchInterface $route_match) 
+{
```
kalpaitch: Interestingly Drupal coding standards don't follow PSR2 yet. I reckon it's probably better to stick with one set of coding standards for now... chat about it next week.

stemount:
thx, yes, this will become a psuedo-field as per prototypes branch. thx

stemount:
This PR is really blocked because this is not featurable. @todo investigate and create solution reusable pseudo field or something that exposes it to views + display + also is featureable

PR #185, opened 2017-08-15
Adds document icon to .file class

nao360:
Icon is already part of assets/images, so this just adds some styles to the .file class so that the document icon is displayed whenever the .file class is used. We can get more specific with the selector (if necessary), but this seems to work well for now.

image


web/themes/custom/par_theme/sass/par.scss
```diff
@@ -250,11 +250,17 @@ a.no-pointer-events {
...
   border-bottom: 1px solid #bfc1c3;
   margin-bottom: 3em;
 }
+
+/* Iconography */
+
+.file {
+  background: url(/themes/custom/par_theme/assets/images/document-icon.png) 0px 2px/15px 20px no-repeat;
+  padding-left: 20px;
+}
```
stemount: Cool.

no new line at EOF… not super important but %extend may make life easier.

PR #186, opened 2017-08-15
Wraps .table-scroll-wrapper in .form-group class

nao360:
Provides bottom margin of 30px. Elements rendered after tables are spaced correctly below the table.

image


stemount:
@nao360 check out ParFlowTransitionAdviceListController.php for an alternative.

`    // Show the documents in table format.
    $build['documentation_list'] = [
      '#theme' => 'table',
      '#attributes' => ['class' => 'form-group'],
      '#title' => 'Advice documentation',
      '#header' => ['Document', 'Type of document and regulatory functions', 'Actions', 'Confirmed'],
      '#empty' => $this->t("There is no documentation for this partnership."),
    ];`

nao360:
@stemount - cheers buddy, I'm thinking the wrapper in the markup is advantageous given that we'll always need that margin under tables. Happy to pass it in as an attribute from the controller if you prefer, but I was thinking this way there's one less parameter we've got to pass in.

nao360:
@kalpaitch / @stemount  ParFlowTransitionInspectionPlanForm doesn't have a controller (I suspect because there's only ever one inspection plan to list). This will have to be one for tomorrow, there's no way I'm venturing into refactor-land without a fresh pair of eyes! 

Let's discuss - I still think there's a case to be made for changing the template - from my POV the template is missing the .form-group wrapper to begin with. Alternative would be to extend .table-scroll-wrapper SASS, and give it the bottom margin property it should have.

kalpaitch:
So adding .form-group to the table template assumes that all tables are forms... there are a number of cases where this isn't the case.

nao360:
Reverted table template changes (no more .form-group div wrapper), and updated form controller to pass .form-group class via #attributes property.

image

web/modules/custom/par_flow_transition_partnership_details/src/Controller/ParFlowTransitionAdviceListController.php
```diff
@@ -49,7 +49,7 @@ public function content(ParDataPartnership $par_data_partnership = NULL) {
     // Show the documents in table format.
     $build['documentation_list'] = [
       '#theme' => 'table',
-      '#attributes' => ['class' => 'form-group'],
+      '#attributes' => ['class' => array('form-group')],
```
stemount: just to keep things uniform - use [] instead of array()

'#attributes' => ['class' => ['form-group']],

web/modules/custom/par_flow_transition_partnership_details/src/Form/ParFlowTransitionInspectionPlanForm.php
```diff
@@ -54,7 +54,8 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
 
     // Show the task links in table format.
     $form['document_list'] = [
-      '#type' => 'table',
+      '#theme' => 'table',
+      '#attributes' => ['class' => array('form-group')],
```
stemount: just to keep things uniform - use [] instead of array()

'#attributes' => ['class' => ['form-group']],

web/themes/custom/par_theme/templates/table.html.twig
```diff
@@ -102,4 +102,4 @@
       </tfoot>
     {% endif %}
   </table>
-</div>
+</div>
```
stemount: no end of line.

web/modules/custom/par_flow_transition_partnership_details/src/Form/ParFlowTransitionInspectionPlanForm.php
nao360: Gaaaaaaaaaaaaaaadddddamnit!

kalpaitch:
Corrected those @stemount

web/themes/custom/par_theme/templates/table.html.twig
nao360: Curious, there's a clear EOL in mine - just added a second one and pushed the commit. Maybe my git CR/LF config?

PR #195, opened 2017-08-16
Changes the error message, 'Sorry dude, this field aint configured fo…

nao360:
Changes the error message, 'Sorry dude, this field aint configured for that', to 'Unknown value' for Transition. Too risky to leave it in.


kalpaitch:
Yup this is safe, no tests for that value. Obviously.

PR #198, opened 2017-08-16
Fixes edit about partnership

nao360:
Updates heading, and hint text on 'About partnership' edit page.

The page heading h1 tag needs a '.heading-large' class passed to page-title template. @stephenb01 or @stemount, I think we'll need to pair for this - t'is not immediately clear how to pass the attribute to the page title template as I can only find YML to work with.


nao360:
NOTE: WE'RE GOING TO SCRAP THIS

Just keeping it here for reference until the changes are made in the DB

PR #199, opened 2017-08-16
Access

kalpaitch:
@stemount this is what you want... `$memberships = $par_data_manager->hasMemberships($account, 'par_data_partnerships');`

That will get you an array keyed by entity id of all entities the user account has access to. So if you just want the ids, which I think you do, you'll want to `array_key()` the result or venture into stack overflow territory.

I created a user account for `user+3034@localhost.localdomain` (who has access to 3 authorities and about 80+ partnerships).The numbers could get reasonably large (100 or so), but hey it's drupal sql, it's huge anyway.


web/modules/custom/par_data/src/Plugin/views/filter/ParMember.php
```diff
@@ -0,0 +1,56 @@
...
+//    $account = User::load(\Drupal::currentUser()->id());
+
+    // Find memberships.
+//    $this->par_data_manager->hasMemberships($account);
+
+    // @todo try get the entity type of the view.
+    $par_data_partnership_type = $this->par_data_manager->getParEntityType('par_data_partnership');
+
+    // Get revision table e.g. "par_partnerships_field_revision.id".
+    $revision_table = $par_data_partnership_type->getRevisionDataTable() . '.id';
```
kalpaitch: Technically revision field :)

stemount: hah. Aye :)

PR #217, opened 2017-08-19
Cucumber

jaffamonkey:
Added rudimentary search results check (easily expanded on, once more varied data available), and generally removing dependencies on specific partnerships and businesses.


kalpaitch:
We *REALLY* need to turn stop using the par_data_test module now. Unfortunately we're still using the users created in this module: `par_authority@example.com` and `par_helpdesk@example.com` etc

Suggestions on how to get a user to login as please. Perhaps we can pick a few examples of the key *REAL* users in the migrated data to test against.

stemount:
Fixing the UUIDs is cool.

PR #221, opened 2017-08-20
Deployment script and Opcache

chris-moreton:
Various changes to the push.sh script. Also enabled opcache.


.bp-config/php/php.ini
```diff
@@ -0,0 +1,1971 @@
...
+; "auto" detect order is changed accoding to mbstring.language
+; http://php.net/mbstring.detect-order
+;mbstring.detect_order = auto
+
+; substitute_character used when character cannot be converted
+; one from another
+; http://php.net/mbstring.substitute-character
+;mbstring.substitute_character = none
+
+; overload(replace) single byte functio
```
kalpaitch: Would be good to explicitly enable this line please.

kalpaitch: And this line please `opcache.load_comments=1`

kalpaitch: Think it would probably be good to double the default `opcache.max_accelerated_files=4000`

kalpaitch: What do you reckon to doubling the defaults `opcache.memory_consumption=128`??

kalpaitch: Think also for this... `opcache.interned_strings_buffer=8`

chris-moreton: Yep - will do

chris-moreton: No problem

PR #228, opened 2017-08-21
Access fixes

kalpaitch:


web/modules/custom/par_roles/par_roles.module
```diff
@@ -23,40 +23,26 @@ function par_roles_help($route_name, RouteMatchInterface $route_match) {
   }
 }
```
jon-edery: Nice to have:  implements hook_... comment

kalpaitch: yup thanks, will add once we get this in and tagged.

PR #229, opened 2017-08-21
PAR-279: Updated the user save routine.

stephenb01:


web/modules/custom/par_flow_transition_business/src/Form/ParFlowTransitionDetailsForm.php
```diff
@@ -148,8 +148,7 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
     // Primary contact summary.
     $par_data_contacts = $par_data_partnership->getOrganisationPeople();
     $par_data_primary_person = array_shift($par_data_contacts);
-
-    $primary_person_view_builder = $this->getParDataManager()->getViewBuilder('par_data_person');
+    $primary_person_view_builder = $par_data_primary_person->getViewBuilder();
```
stemount: this has regressed, the above is right.

stephenb01: darn I missed one!!!

web/modules/custom/par_flow_transition_business/src/Form/ParFlowTransitionTermsForm.php
```diff
@@ -66,7 +66,7 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
     // Partnership agree terms.
     $form['terms_conditions'] = [
       '#type' => 'checkbox',
-      '#title' => t('I confirm that my organisation agrees to the new terms and conditions.'),
+      '#title' => t('I confirm that my organisation agrees to the new <a href="\terms-and-conditions">terms and conditions.</a>'),
```
kalpaitch: @nao360 what was the verdict on checkbox labels, we should add this as a separate piece of text no.

web/modules/custom/par_flow_transition_partnership_details/src/Form/ParFlowTransitionTermsForm.php
```diff
@@ -66,7 +66,7 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
     // Partnership details.
     $form['terms_conditions'] = [
       '#type' => 'checkbox',
-      '#title' => t('I confirm that my authority agrees to the new terms and conditions.'),
+      '#title' => t('I confirm that my authority agrees to the new <a href="\terms-and-conditions">terms and conditions.</a>'),
```
kalpaitch: Same here.

web/modules/custom/par_flow_transition_business/src/Form/ParFlowTransitionTermsForm.php
kalpaitch: Maybe we could put it on the '#description' key instead??

nao360:
@stephenb01 a thousand apologies, I forgot to update the screenshot you worked against! We're not adding the hyperlink to the checkbox label, we're adding the hyperlink to the text above it.

Illustration below:

screen shot 2017-08-20 at 13 43 17

web/modules/custom/par_flow_transition_business/src/Form/ParFlowTransitionDetailsForm.php
```diff
@@ -234,19 +233,19 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
...
-
+        $form['legal_entity'][$legal_entity_item->id()] = [
+          '#type' => 'fieldset',
+          '#title' => '',
+          '#collapsible' => FALSE,
+          '#collapsed' => FALSE,
+        ];
         $alternative_legal = $legal_entity_view_builder->view($legal_entity_item, 'full');
-        $form['legal_entity']['alternative'][$legal_entity_item->id()] = $this->renderMarkupField($alternative_legal);
+        $form['legal_entity'][$legal_entity_item->id()]['value'] = $this->renderMarkupField($alternative_legal);
```
stemount: `['value']` is not needed, should renderMarkupField get used to actually set some other FAPI settings in the future. best to keep this uniform.

stephenb01: Actually we do need the ['value'] here but it doesn't have to be called value. Without it, the fieldset above will be overwritten. I have replaced this with a better solution - that actually does what I want.

PR #232, opened 2017-08-22
Access fixes: Updated permissions checks for Bug #19

kalpaitch:


nao360:
@kalpaitch tested, and I'm getting a PHP error as follows:


Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /var/www/html/web/core/lib/Drupal/Core/Entity/ContentEntityBase.php on line 207

Can test with php mem limit set to 256M - please advise.

nao360:
Passes with PHP memory limit set to 256M

PR #250, opened 2017-08-23
PAR-537: Fixed bug in getting hte organisation in the partnership det…

stephenb01:
…ails.


web/modules/custom/par_flow_transition_partnership_details/src/Form/ParFlowTransitionOverviewForm.php
```diff
@@ -157,7 +157,7 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
     $organisation_primary_person = array_shift($organisation_people);
 
     // List the Primary Organisation contact.
-    $par_data_organisation = current($par_data_partnership->retrieveEntityValue('par_data_organisation'));
+    $par_data_organisation = current($par_data_partnership->retrieveEntityValue('field_organisation'));
```
stemount: how about `$par_data_organisation = $par_data_partnership->retrieveEntityValue('field_organisation');`, does that work? `current()` should not be necessary here.

stephenb01: It is required here. Without it we get an array of organisations (even if there is only one).

kalpaitch: current() is required

PR #257, opened 2017-08-23
Allow CDN routes for non-prod envs

chris-moreton:


web/sites/default/settings.php
```diff
@@ -743,7 +743,8 @@
 $appEnv = getenv('APP_ENV');
 
 $settings['trusted_host_patterns'] = [
-  '^par-beta-' . $appEnv . '\.cloudapps\.digital',
+ '^par-beta-' . $appEnv . '\.cloudapps\.digital',
```
stemount: there is only one space here instead of 2.

PR #259, opened 2017-08-25
PAR S3 file download support

jon-edery:
End to end tests complete confirmed patches applied correctly and code updates resolve the download S3 files issue.

This PR provides support for private S3 stored files to be download via Drupal 8.
The file entity contrib module does not support this functionality although a workaround exists for Drupal 7.

These patches invoke a new hook that provides modules with the ability to request that a file is treated as an s3 based file.  A new response object class is also provided to deal with S3 based files.


composer.json
```diff
@@ -108,6 +108,9 @@
             "drupal/flysystem_s3": {
                 "Allow downloading of private files from remote S3 wrappers": "./patches/private_file_download.patch"
             },
+          "drupal/file_entity": {
```
stemount: couple of extra spaces

jon-edery:
tested and ready to merge.

PR #266, opened 2017-08-31
ClamAV scanning of files via REST daemon

stemount:
patch ClamAV to scan files on upload.
two docker boxes, one a REST client, one a clamav tcp daemon.
configuration for ClamAV module to use REST service.


patches/clamav_add_rest_client.patch
```diff
@@ -0,0 +1,242 @@
...
++   * {@inheritdoc}
++   */
++  public function version() {
++
++    try {
++
++      $request = $this->_httpclient->get($this->_endpoint);
++
++      if ($json = json_decode($request->getBody())) {
++        return isset($json->version) ?: 'unknown';
```
kalpaitch: This will return the value of isset() not the version itself...

patches/clamav_add_rest_client.patch
```diff
@@ -0,0 +1,242 @@
...
++
++namespace Drupal\clamav\Scanner;
++
++use Drupal\file\FileInterface;
++use Drupal\clamav\ScannerInterface;
++use Drupal\clamav\Scanner;
++use Drupal\clamav\Config;
++
++class DaemonRestClient implements ScannerInterface {
++  protected $_file;
```
kalpaitch: Why leading underscore?? http://www.php-fig.org/psr/psr-1/#42-properties

patches/clamav_add_rest_client.patch
```diff
@@ -0,0 +1,242 @@
...
++  public function scan(FileInterface $file) {
++
++    $result = Scanner::FILE_IS_UNCHECKED;
++
++    try {
++      $file_post = $this->_httpclient->post($this->_endpoint, [
++        'multipart' => [
++          [
++            'name' => 'file',
++            'contents' => fopen($file->getFileUri(), 'r')
```
kalpaitch: Hmm, so our endpoint isn't authenticated, which means we're running an open anti-virus scanning service??

patches/clamav_add_rest_client.patch
```diff
@@ -0,0 +1,242 @@
...
++          ],
++          [
++            'name' => 'name',
++            'contents' => $file->getFilename()
```
kalpaitch: Is there any reason why we couldn't send the filename as `filename` on the first portion of the multipart?

kalpaitch:
Cool, looks awesome, some minor nit picking on the patch which really doesn't make much difference. Also I'm gonna just have a little look at the ClamAV module in more depth just so I understand what it's doing and how asynchronous the process really is :)

patches/clamav_add_rest_client.patch
stemount: "Whatever naming convention is used SHOULD be applied consistently within a reasonable scope."

Not sure if this applies, but the other scanners in the folder seem to follow this.

patches/clamav_add_rest_client.patch
stemount: The endpoint will be authenticated with an API key as discussed with @chris-moreton

kalpaitch: Yea cool, I just couldn't see any mention of API key in this PR, might have missed it..

patches/clamav_add_rest_client.patch
kalpaitch: It's not important, but kinda case in point, in my mind the scope for Drupal projects is Drupal, and it looks like the ClamAV module isn't consistent with this, not your boggle though.

patches/clamav_add_rest_client.patch
stemount: @kalpaitch this is now in!

PR #267, opened 2017-08-31
Update deploy script to perform blue/green deploys

chris-moreton:


kalpaitch:
Very cool, just a q though, does this mean we can only use this push scripts to envs with a CDN?

PR #272, opened 2017-09-01
PAR-338: Added new view for listing partnerships.

stephenb01:
- View name par_data_beta_partnerships
- View URL /partnerships
- View code in module. Might be better in a theme. Suggestions!


web/modules/features/par_partnership_flows/par_partnership_flows.module
```diff
@@ -0,0 +1,48 @@
...
+ * Alter views for the the beta journeys.
+ */
+
+use Drupal\views\ViewExecutable;
+
+/**
+ * Implements hook_views_pre_view().
+ */
+function par_partnership_flows_views_pre_view(ViewExecutable $view, $display_id, array &$args) {
+  // @todo This routine probably needs to be put into a module or theme.
```
kalpaitch: Agreed, I think it's probably worth doing this view properly like you suggest, I can see you're trying to get this in so that a full flow can be started, but I think it's worth doing this properly, there's a massive risk atm that if we do temporary code we'll just not have time to replace it and leave it in forever.

kalpaitch:
Just a couple of thoughts:
I don't think this view relates to the par_data model, nor do I think that any of the code we're writing atm is just related to 'beta' so might be better to call it `par_partnership` or `par_my_partnerships`???

web/modules/features/par_partnership_flows/par_partnership_flows.module
stephenb01: We still need to use the hook_views_pre_view() in any case don't we?

stemount: @kalpaitch @stephenb01 how about a custom block that is set to show on the top of certain URLs instead?

stephenb01: getting confused on how this should be implemented. :(

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAuthorityDetailsForm.php
```diff
@@ -251,35 +220,196 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
...
+      '#type' => 'fieldset',
+      '#title' => t('Advice and Documents'),
+      '#attributes' => ['class' => 'form-group'],
+      '#collapsible' => FALSE,
+      '#collapsed' => FALSE,
     ];
+    $form['advice']['edit'] = [
+      '#type' => 'markup',
+      '#markup' => t('@link', [
+        '@link' => $this->getFlow()->getNextLink('advice')->setText('See all Advicet')->toString(),
```
stemount: see all advicet :-)

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAuthorityDetailsForm.php
```diff
@@ -63,130 +63,99 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
...
+      '#collapsed' => FALSE,
+    ];
+    if ($par_data_organisation->get('comments')->getString() === '') {
+      $form['about_business']['info'] = [
+        '#type' => 'markup',
+        '#markup' => $this->t('(none)'),
+      ];
+    }
+    else {
+      $form['about_business']['info'] = $this->renderMarkupField($about_organisation);
```
stemount: renderMarkupField does the (none) thing for you. `$form['about_business']['info'] = $this->renderMarkupField($about_organisation);` would suffice here

stephenb01: Just tried this again and it doesn't show (none) value.

stephenb01: I see I need to pass an empty value instead of about_organisation to get the (none) to appear.
This will probably always return a value for $about_organisation:
`$about_organisation = $par_data_organisation ? $organisation_builder->view($par_data_organisation, 'about') : '';`

stemount: yea, try `$about_organisation = $organisation_builder->view($par_data_organisation, 'about');`

stephenb01: That doesn't work as $about_organisation has something in it.

stephenb01: This works:
`  public function renderMarkupField($field) {
    $rendered_field = $this->getRenderer()->render($field);
    return [
      '#type' => 'markup',
      '#markup' => $rendered_field ? $rendered_field : '(none)',
    ];
  }`

web/modules/features/par_partnership_flows/par_partnership_flows.routing.yml
```diff
@@ -3,9 +3,9 @@ par_partnership_flows.authority_details:
   path: '/partnership/{par_data_partnership}/authority-details'
   defaults:
     _form: '\Drupal\par_partnership_flows\Form\ParPartnershipFlowsAuthorityDetailsForm'
-    _title: 'Partnership Details for Primary Authority'
+    _title: ''
```
stemount: is this intentional?

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAuthorityDetailsForm.php
```diff
@@ -63,130 +63,93 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
...
-          '#attributes' => ['class' => 'form-group'],
-          '#collapsible' => FALSE,
-          '#collapsed' => FALSE,
-        ];
+    $form['about_business']['info'] = $this->renderMarkupField($about_organisation);
 
-        $alternative_person = $person_view_builder->view($premises, 'full');
-        $form['registered_address'][$premises->id()]['premises'] = $this->renderMarkupField($alternative_person);
+    // Sic Codes.
+    $par_data_sic_code = $par_data_organisation->getSicCode();
```
stemount: do we need getters for everything? could this just be $blah->get('field_sic_code')->referencedEntities();

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAuthorityDetailsForm.php
```diff
@@ -63,130 +63,93 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
...
+    $form['sic_codes'] = [
+      '#type' => 'fieldset',
+      '#title' => t('SIC Code'),
```
stemount: do these `#title`s need a colon at the end to be consistent?

PR #280, opened 2017-09-03
PAR-338: Updated the about partnership form for beta.

stephenb01:


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAboutForm.php
```diff
@@ -76,51 +76,45 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
...
+      'comments' => [
+        'value' => $form_state->getValue('about_partnership'),
+        'key' => 'about_partnership',
+        'tokens' => [
+          '%field' => !empty($form['about_partnership']['#title']) ? $form['about_partnership']['#title']->render() : '',
+        ]
+      ],
+    ];
+
+    $errors = $par_data_partnership->validateFields($fields);
```
kalpaitch: Awesome, if we're starting to think about validation for Beta I think we should start thinking about transplanting this functionality to the [Validation API](https://www.drupal.org/docs/8/api/entity-validation-api/entity-validation-api-overview), don't think we want to be maintaining our own way of doing this...

Example here https://github.com/TransformCore/beis-par-beta/blob/master/web/modules/custom/par_flows/modules/par_demo_forms/src/Form/ParDemoFirstForm.php#L70

Can get this in for now but would suggest we don't do any more validation without this otherwise we'll be duplicating effort.

stephenb01: Will look into it. I agree, stat as you mean to go on.

web/modules/custom/par_data/src/Entity/ParDataPartnership.php
```diff
@@ -250,6 +250,7 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
       ->setLabel(t('About Partnership'))
       ->setDescription(t('Details about this partnership.'))
       ->setRevisionable(TRUE)
+      ->setRequired(TRUE)
```
kalpaitch: Just wondering where you got that these fields are required?

web/modules/custom/par_flows/src/Form/ParBaseForm.php
```diff
@@ -267,6 +295,21 @@ public function buildForm(array $form, FormStateInterface $form_state) {
    */
   public function validateForm(array &$form, FormStateInterface $form_state) {
 
+    foreach ($this->getformItems() as $entity_name => $form_items) {
+      $entity_type_def = $this->getParDataManager()->getParEntityType($entity_name);
+      $entity = $entity_type_def->getClass()::create([
+        'type' => $this->getParDataManager()->getParBundleEntity($entity_name)->id(),
```
kalpaitch: Just to highlight that this will only work if there is only one bundle for the entity, this is not true for organisations atm, and possibly to be true for partnerships.

web/modules/custom/par_data/src/Entity/ParDataPartnership.php
stephenb01: Yeah I needed something to test against. But like you said we do need confirmation for Beta phase on what is required and what is not. Perhaps @nao360 needs to add this to the jira tickets so Paul can test it fully.

web/modules/custom/par_flows/src/Form/ParBaseForm.php
stephenb01: Yeah realised that when I added it in. Happy to do what you mentioned above.

PR #293, opened 2017-09-05
PAR-557: Updated physical model. Added new fields only

kalpaitch:
No fields updated or removed.

Added new entity:
**coordinated_business** (often spelt as coordianted, coodinated or a mix of the two)

Added new fields only:
**revoked**: partnership, inspection_plan, advice
**issue_date**: advice
**field_person**: enforcement_notice
**field_regulatory_function**: inspection_plan


web/modules/custom/par_validation/src/Plugin/Validation/Constraint/ParRequiredValidator.php
```diff
@@ -0,0 +1,37 @@
...
+      }
+    }
+  }
+
+  /**
+   * Determine the contexts under which this validation is appropriate.
+   */
+  public function validationIsAllowed() {
+    $current_user = \Drupal::currentUser();
+    return $current_user->hasPermission('par bypass validation') ? FALSE : TRUE;
```
stemount: is this `false : true`?

kalpaitch: Updated.

web/modules/custom/par_data/src/Entity/ParDataOrganisation.php
```diff
@@ -253,6 +253,49 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
+        'type' => 'string_textfield',
+        'weight' => 1,
+      ])
+      ->setDisplayConfigurable('form', FALSE)
+      ->setDisplayOptions('view', [
+        'label' => 'hidden',
+      ])
+      ->setDisplayConfigurable('view', TRUE);
+
+    // Coordinator type.
```
stephenb01: Label incorrect...

PR #296, opened 2017-09-06
PAR-335: Initial set up of dashboard.

stephenb01:
- Very basic page with routing. 
- Need to work out permissions.

### DIFF
--- a/sync/config_split.config_split.dev_config.yml
+++ b/sync/config_split.config_split.dev_config.yml
@@ -15,7 +15,8 @@ module:
 theme: {  }
 blacklist:
   - par_data.settings
-graylist: {  }
-graylist_dependents: true
+graylist:
+  - login_destination.destination.welcome_reminder
+graylist_dependents: false
 graylist_skip_equal: true
 weight: 0
--- a/sync/config_split.config_split.dev_config.yml
+++ b/sync/config_split.config_split.dev_config.yml
@@ -15,8 +15,7 @@ module:
 theme: {  }
 blacklist:
   - par_data.settings
-graylist:
-  - login_destination.destination.welcome_reminder
-graylist_dependents: false
+graylist: {  }
+graylist_dependents: true
 graylist_skip_equal: true
 weight: 0
--- a/sync/user.role.par_business.yml
+++ b/sync/user.role.par_business.yml
@@ -42,6 +42,7 @@ permissions:
   - 'edit par_data_regulatory_function entities'
   - 'edit par_data_sic_code entities'
   - invite_type_invite_organisation_member
+  - 'manage my organisations'
   - 'use text format basic_html'
   - 'view own private files'
   - 'view unpublished par_data_advice entities'
--- a/sync/user.role.par_business.yml
+++ b/sync/user.role.par_business.yml
@@ -20,6 +20,7 @@ permissions:
   - 'access par_data_regulatory_function entities'
   - 'access par_data_sic_code entities'
   - 'access partnerships dashboard'
+  - 'busines partnership journey'
   - 'create files'
   - 'delete own document files'
   - 'delete own image files'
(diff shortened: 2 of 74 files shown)


kalpaitch:
I'll fix the tests in master then we can get this running. Can I be a pain though and question where this controller goes, my feeling is that we have a par_dashboards module already, and this functionality isn't related much to 'login'...

web/modules/features/par_dashboards/par_dashboards.routing.yml
```diff
@@ -0,0 +1,13 @@
+# PAR Login Dashboard
+par_login.dashboard:
```
kalpaitch: Might want to change the module name in this config :)

stephenb01: Arrgghhh!!! :(

sync/user.role.par_business.yml
```diff
@@ -20,6 +20,7 @@ permissions:
   - 'access par_data_regulatory_function entities'
   - 'access par_data_sic_code entities'
   - 'access partnerships dashboard'
+  - 'busines partnership journey'
```
kalpaitch: sp

stephenb01: Actually that is how it is!
Need to alter the permission!

sync/config_split.config_split.dev_config.yml
```diff
@@ -15,7 +15,8 @@ module:
 theme: {  }
 blacklist:
   - par_data.settings
-graylist: {  }
-graylist_dependents: true
+graylist:
```
kalpaitch: I don't think this should be here. It's meant to be overridden in the `beta_config` no??

sync/user.role.par_business.yml
kalpaitch: Oops, that must defo be one of mine. Let's leave for now and we'll come back to permissions later.

sync/config_split.config_split.dev_config.yml
stephenb01: This is for the dev config so it can be tested. Wasn't sure if I should have this here.

web/modules/features/par_dashboards/src/Controller/ParDashboardsDashboardController.php
```diff
@@ -0,0 +1,80 @@
...
+    $account = User::load(\Drupal::currentUser()->id());
+
+    $build['partnerships'] = [
+      '#type' => 'fieldset',
+      '#attributes' => ['class' => 'form-group'],
+      '#collapsible' => FALSE,
+      '#collapsed' => FALSE,
+    ];
+
+    if ($account->hasPermission('bypass partnership journey')) {
```
kalpaitch: You don't need to to user load for this, you can just do `\Drupal::currentUser()->hasPermissions('bypass partnership journey')` Less costly.

web/modules/features/par_dashboards/src/Controller/ParDashboardsDashboardController.php
```diff
@@ -14,11 +13,15 @@ class ParDashboardsDashboardController extends ParBaseController {
...
    */
   public function content() {
-
-    $build['intro'] = [
-      '#type' => 'markup',
-      '#markup' => $this->t('Intro to new dashboard'),
-    ];
+    $current_user = \Drupal::currentUser();
+    $roles = $current_user->getRoles();
+    if (in_array('par_authority', $roles)) {
```
kalpaitch: Best not to rely on roles, should always do a permission check (knowing that that permission is assigned to that, and only that role). We may have put other ones like that in transition, if you see any make the switch. We'll need to think about what our real permissions are for beta too at some point, maybe next week though.

stephenb01: Yeah did wonder if this was right or not.

PR #304, opened 2017-09-08
Par 583

stephenb01:


sync/core.entity_view_display.par_data_organisation.business.name.yml
```diff
@@ -0,0 +1,78 @@
+uuid: ad64136a-bffd-4458-93d0-76b44bffd14e
```
kalpaitch: Maybe have a look to see if we can't update the `core.entity_view_display.par_data_organisation.business.title.yml` instead, perhaps see if it's being used anywhere else.

Either way need to update the par_data_config feature.

PR #309, opened 2017-09-11
Sql report

chris-moreton:


stemount:
@chris-moreton is this just something as a temporary measure? can the RD dashboard be improved with your changes?

PR #313, opened 2017-09-11
Password strength

chris-moreton:
I have installed the module plugin "password_strength" for the module "password_policy". I removed all the matchers except for the dictionary lookup matcher. I created a patch which changes the message to be specific about the dictionary matching.


kalpaitch:
Awesome stuff, unfortunately because we've done the invitation/registration form in a weird way it actually bypasses the password policy checks completely. Nothing to do with this PR, but until we update the invite/registration form this won't do much useful checking...

stemount:
In future - We could change the messaging using t() string overrides over a patch.

stemount:
@chris-moreton If it works on user/%/edit, then this is a winner. our invite custom module can just invoke this I guess.

PR #314, opened 2017-09-12
Par 370

stephenb01:


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsOrganisationDetailsForm.php
```diff
@@ -386,22 +386,31 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
...
+        '#type' => 'fieldset',
+        '#title' => t('SIC Code:'),
+        '#collapsible' => FALSE,
+        '#collapsed' => FALSE,
+      ];
+
+      $sic_codes = $par_data_organisation->getSicCode();
+
+      foreach ($sic_codes as $key => $sic_code) {
+        $id = $sic_code->id();
```
jon-edery: Can id() = NULL?  should we be checking the value before passing it to an array element?

stephenb01: :)

stephenb01: Should never be NULL but I will put a check in just in case!

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsOrganisationDetailsForm.php
```diff
@@ -386,22 +386,36 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
...
-    $form['advice']['edit'] = [
-      '#type' => 'markup',
-      '#markup' => t('@link', [
-        '@link' => $this->getFlow()
-          ->getNextLink('advice')
-          ->setText('See all Advice')
-          ->toString(),
-      ]),
-    ];
+    if ($this->getFlowName() === 'partnership_direct') {
```
kalpaitch: let's use `$par_data_partnership->isDirect()` instead to check this please.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsSicCodeForm.php
```diff
@@ -44,20 +42,28 @@ public function retrieveEditableValues(ParDataPartnership $par_data_partnership
   /**
    * {@inheritdoc}
    */
-  public function buildForm(array $form, FormStateInterface $form_state, ParDataPartnership $par_data_partnership = NULL) {
+  public function buildForm(array $form, FormStateInterface $form_state, ParDataPartnership $par_data_partnership = NULL, $sic_code_delta = NULL) {
     $this->retrieveEditableValues($par_data_partnership);
     $par_data_organisation = current($par_data_partnership->getOrganisation());
```
kalpaitch: It would be great to see if we can't do `$par_data_partnership->getOrganisation()->first()` instead of `currenty()` here as discussed.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsSicCodeForm.php
```diff
@@ -44,20 +42,28 @@ public function retrieveEditableValues(ParDataPartnership $par_data_partnership
...
 
     $form['intro'] = [
       '#markup' => $this->t('Change the SIC Code of your business'),
     ];
 
+    $options = [];
+
+    // Get the list of valid sic codes.
+    // Probably not the best method, but will do for now!
+    $sic_codes = \Drupal::entityTypeManager()->getStorage('par_data_sic_code')->loadMultiple();
```
kalpaitch: Can we use `$this->getParDataManager()` methods rather than global scope stuff. Shouldn't need to use \Drupal::static... we can adapt the par data manager if it doesn't yet fit our needs.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsSicCodeForm.php
stephenb01: This doesn't appear to work using: $par_data_partnership->getOrganisation()->first()

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsSicCodeForm.php
```diff
@@ -44,20 +42,27 @@ public function retrieveEditableValues(ParDataPartnership $par_data_partnership
...
+    }
 
     $form['sic_code'] = [
-      '#type' => 'textfield',
+      '#type' => 'select',
       '#title' => $this->t('SIC Code'),
-//      '#default_value' => isset($par_data_organisation->get('trading_name')->getValue()[$trading_name_delta]) ? $par_data_organisation->get('trading_name')->getValue()[$trading_name_delta] : '',
-//      '#description' => $this->t('Sometimes companies trade under a different name to their registered, legal name. This is known as a \'trading name\'. State any trading names used by the business.'),
+      '#options' => $options,
+      '#default_value' => isset($par_data_organisation->get('field_sic_code')->getValue()[$sic_code_delta]) ? $par_data_organisation->get('field_sic_code')->getValue()[$sic_code_delta] : '',
```
kalpaitch: Should have spotted this before, we can't use $sic_code_delta here because we have to assume it's adding a new one and doesn't specify the delta.

As discussed this morning we should check all forms as we go to make sure we only use entities in the `retrieveEditableValues()` method, if we want to their values in the main form it's better to do `getDefaultValues()` which makes sure to retrieve previous form sessions too.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsSicCodeForm.php
kalpaitch: ++

PR #315, opened 2017-09-12
PAR-619: Fixed issue with permission for coordinated journey.

stephenb01:


kalpaitch:
@stephenb01 we can merge if you'd like and I can look at all permissions in PAR-617 if you'd like??

PR #316, opened 2017-09-12
PAR-620: Fixed issue employee number not showing (none).

stephenb01:


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsOrganisationDetailsForm.php
```diff
@@ -173,7 +173,7 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
       '#collapsed' => FALSE,
     ];
 
-    if ($par_data_organisation->get('employees_band')->getString() !== '0') {
+    if ($par_data_organisation->get('employees_band')->getString() !== '') {
```
jon-edery: FALSE would be a more robust option here.

stephenb01: It needs to be '' otherwise it doesn't work.

kalpaitch: Maybe `!$par_data_organisation->get('employees_band')->isEmpty()`

jon-edery: empty string should evaluate to FALSE:  

 When converting to boolean, the following values are considered FALSE as well as the following:

    the boolean FALSE itself
    the integer 0 (zero)
    the float 0.0 (zero)
    the empty string, and the string "0"
    an array with zero elements
    the special type NULL (including unset variables)
    SimpleXML objects created from empty tags

PR #335, opened 2017-09-14
PAR-399: Added advice list and inspection lists to organisation and enforcement journeys.

kalpaitch:


web/modules/features/par_partnership_flows/par_partnership_flows.routing.yml
```diff
@@ -196,7 +197,7 @@ par_partnership_flows.address_add:
       par_data_partnership:
         type: entity:par_data_partnership
 par_partnership_flows.organisation_contact_edit:
-  path: '/partnership/{par_data_partnership}/organisation/{par_data_organisation}/contact/{par_data_person}'
+  path: '/partnership/{par_data_partnership}/organisation/contact/{par_data_person}'
```
kalpaitch: @stemount I had to change this back to how it was because of the way we are now adding 'sections'

PR #347, opened 2017-09-15
PAR-648: Added code to manage field lookup items.

stephenb01:


kalpaitch:
Hmm, let us bang our heads together until the blood rushes out. Then We'll know what to do with this..

PR #353, opened 2017-09-15
multiset form config for search par redirect button

jon-edery:
Fix for the done button on the par flow multi-set form. config added to set redirect route for partnership/PAR_ID pages to redirect to the partnership search page.


kalpaitch:
Seems like config be outta sink with features is all.

PR #354, opened 2017-09-15
PAR-650: Remove links from dashboard.

stephenb01:


kalpaitch:
Think there's one more to remove which is the 'Send enforcement notice...' or whatever on the partnerships page when you click through from search.

PR #366, opened 2017-09-17
Cancel all flow forms and cancel flow form submit handlers.

stemount:
An option for previous/back and flow cancel.

example usage
`
$form['actions']['cancel'] = [
      '#type' => 'submit',
      '#name' => 'cancel',
      '#value' => $this->t('Cancel'),
      '#submit' => ['::cancelForm'],
      '#limit_validation_errors' => [],
      '#attributes' => [
        'class' => ['btn-link']
      ],
    ];
`

### DIFF
--- a/web/modules/custom/par_flows/src/Form/ParBaseForm.php
+++ b/web/modules/custom/par_flows/src/Form/ParBaseForm.php
@@ -372,6 +372,54 @@ abstract class ParBaseForm extends FormBase implements ParBaseInterface {
   }
 
   /**
+   * Cancel submit handler to clear all the current flow temporary form data.
+   *
+   * @code
+   * $form['actions']['cancel'] = [
+   *   '#type' => 'submit',
+   *   '#name' => 'cancel',
+   *   '#value' => $this->t('Cancel'),
+   *   '#submit' => ['::cancelForm'],
+   * ];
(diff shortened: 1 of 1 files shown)


web/modules/custom/par_flows/src/Form/ParBaseForm.php
```diff
@@ -372,6 +372,54 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
+   *
+   * @param array $form
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   */
+  public function cancelForm(array &$form, FormStateInterface $form_state) {
+    // Delete form storage.
+    $this->deleteStore();
+
+    // Go to cancel step.
+    $next = $this->getFlow()->getNextRoute('cancel');
```
kalpaitch: Minor point, usually cancel takes you back a step so I think it might be better to be `getPrevRoute`

stemount: because cancel obliterates the whole flows form data anyway, should we only really make the user go back 1 step?

stephenb01: User needs to go back to the start or landing page.

PR #368, opened 2017-09-17
PAR-419 base form for enforcement notification actions

jon-edery:
image

### DIFF
--- a/web/modules/features/par_enforcement_flows/src/Form/ParEnforcementAddActionForm.php
+++ b/web/modules/features/par_enforcement_flows/src/Form/ParEnforcementAddActionForm.php
@@ -39,7 +39,7 @@ class ParEnforcementAddActionForm extends ParBaseForm {
     $enforcement_notice_bundle = $this->getParDataManager()->getParBundleEntity('par_data_enforcement_notice');
     $reg_function_bundle = $this->getParDataManager()->getParBundleEntity('par_data_regulatory_function');
 
-    $form['action'] =[
+    $form['action'] = [
       '#type' => 'fieldset',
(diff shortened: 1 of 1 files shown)


web/modules/features/par_enforcement_flows/src/Form/ParEnforcementAddActionForm.php
```diff
@@ -36,10 +36,72 @@ public function retrieveEditableValues() {
...
+    ];
+
+    $form['enforcement_type'] = [
+      '#type' => 'radios',
+      '#title' => $this->t('Enforcing type'),
+      '#options' => $enforcement_notice_bundle->getAllowedValues('notice_type'),
+      '#default_value' => $this->getDefaultValues('notice_type'),
+      '#required' => TRUE,
+      '#prefix' => '<div>',
+      '#suffix' => '</div></br>',
```
stemount: `</br>`? :grinning:

kalpaitch:
Looks good @jon-edery. I've noticed two differences between the GDS screens you've been working too and the actual data model/drupal entities.

We'll need to discuss these with @nao360, essentially the `Enforcing type` on the add action form should actually be related to the notice as a whole and not to an individual actions (will have to be moved to the raise a notice form), and the `Address of premises concerned (if applicable)` textarea on the raise a notice form doesn't exist in the data model (will have to be removed).

Also I think this `Copy in another enforcing officer(optional)` which we'd agreed to chop out has remained in the GDS screens. I'm afraid there will be a bit of this because it's been so difficult to keep the UX/wireframes up to date with the latest data model/functionality.

I'd suggest we merge this in and make those changes separately though...

PR #369, opened 2017-09-17
PAR-630: Built out some of the screens, routes (tbc) and flow.

stemount:
Start of the partnership application journey.

image

image

image

image

image

image

### DIFF
--- a/sync/par_flows.par_flow.partnership_application.yml
+++ b/sync/par_flows.par_flow.partnership_application.yml
@@ -1,0 +1,52 @@
+uuid: c1a916b4-3ba9-4e7e-ae6c-304d12fe6841
+langcode: en
+status: true
+dependencies: {  }
+_core:
+  default_config_hash: 06Sx79OH-Q6sfgHhjDCLAlYbNEdGDC6fqRHrov1Eh70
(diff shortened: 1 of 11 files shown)


kalpaitch:
Great, happy to merge this one, just gonna leave two comments here in the mean time:

1) Think we should aim for consistency on the green button text, we're using 'Next', 'Save' (where we're not actually saving anything) and 'Continue'
2) If they choose an existing organisation it should probably skip all the 'organisation details' form parts and go straight to confirmation or whatever is next??

stemount:
@kalpaitch I agree with point 1. We need this consistency to run in the wireframes. I have not gone "off piste" 😀 

2. Yes, organisation details wlil not allow you to change the address. This was in the wireframe, and clarified with @dawn this is to change.

RE Checkboxes/radios . The business rules are to be defined next couple of days. We also need to know if we need to store any of these values asap.

kalpaitch:
@stemount Agreed. Just an FYI though we *can not* add any new fields at this point, it's just too late. so bear that in mind. Anything not in the data model will have to be left out of the form or used just for validation and not stored.

PR #372, opened 2017-09-18
PAR-630: Added about partnership form and other forms to compelte flow.

stephenb01:
Added some forms in just for the flow. Will need to work on these individually once the flow is sorted.
Also changed a few classes so we know what they are in the flow.

image
image

image


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAboutForm.php
```diff
@@ -56,20 +56,24 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
...
       '#name' => 'save',
-      '#value' => $this->t('Save'),
+      '#value' => $this->t($par_data_partnership ? 'Save' : 'Continue'),
     ];
 
-    $cancel_link = $this->getFlow()->getPrevLink('cancel')->setText('Cancel')->toString();
-    $form['cancel'] = [
-      '#type' => 'markup',
-      '#markup' => t('@link', ['@link' => $cancel_link]),
+    $form['actions']['cancel'] = [
```
kalpaitch: This should really be in the base form.

stemount: Defo. On my agenda tomorrow.

kalpaitch: Awesome.

PR #387, opened 2017-09-20
PAR-665: Added CSV Uploaders.

kalpaitch:
Very basic, will need a lot of work but enables us to migrate the first round of members without and provides a starting block for when we have to do it for real.

screenshot from 2017-09-20 10-44-42

screenshot from 2017-09-20 10-44-25

screenshot from 2017-09-20 10-43-53


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsMemberConfirmForm.php
```diff
@@ -0,0 +1,291 @@
...
+        foreach($existing_options as $option) {
+          $option_view = $viewBuilder->view($option, 'title');
+
+          $radio_options[$option->id()] = $this->renderMarkupField($option_view)['#markup'];
+        }
+
+        if (count($radio_options) >= 1) {
+          $form['members'][$i]['existing_organisation'] = [
+            '#type' => 'radios',
+            '#title' => t('A match was found for @name', ['@name' => $name]),
```
kalpaitch: In the long run definitely want to find a way to re-use the de-duping screens that @stemount has put in for the application flow.

stemount: indeed

PR #391, opened 2017-09-20
PAR-645: Updated route for partnership invite.

stephenb01:


web/modules/features/par_partnership_flows/par_partnership_flows.routing.yml
```diff
@@ -485,3 +485,10 @@ par_partnership_flows.partnership_application_details:
     _title_callback: '\Drupal\par_partnership_flows\Form\ParPartnershipFlowsApplicationDetailsForm::titleCallback'
   requirements:
     _permission: 'manage my authorities'
+par_partnership_flows.partnership_application_invite:
+  path: '/partnership/apply/invite'
```
kalpaitch: Hmm, yea you will need to put the user that is being invited (and most likely the partnership as well) into the URL. These are both needed in the form.

stephenb01: Yeah, just spotted that. @stemount and I are looking into the previous step first.

stemount:
*to be reopened when this PR is ready*

PR #392, opened 2017-09-20
Par 666

stephenb01:


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsApplicationOrganisationSearchForm.php
```diff
@@ -26,6 +26,17 @@ public function getFormId() {
   /**
    * {@inheritdoc}
    */
+  public function titleCallback() {
+    if ($this->getFlowName() === 'partnership_application') {
+      return 'New Partnership Application';
```
stemount: this one is a form class specifically for the Application thing, this would not be needed `$this->getFlowName`

stephenb01: You never know!!!

kalpaitch:
Works for me, but would suggest we agree how we're going to manage titles to keep consistency. There are a few areas like this where we're starting to do things in a few different ways.

The `titleCallback` is for titles that do NOT differ by route (as in are the same for multiple pages). There are a couple of instances of these:
- The sitewide default 'Primary Authority Register`
- A flow default (currently we do not have this) might be better to put a 'page_title' property on the flow entity than entering `New Partnership Application` on 10 separate occasions.

IMO any other title can be set with the _title property on the route.

stemount:
So, do we work with `_title` and rely on the routes file in this scenario?

kalpaitch:
Dunno, possibly something like https://github.com/TransformCore/beis-par-beta/pull/396

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAboutForm.php
```diff
@@ -28,6 +28,17 @@ public function getFormId() {
   }
 
   /**
+   * {@inheritdoc}
+   */
+  public function titleCallback() {
+    if ($this->getFlowName() === 'partnership_application') {
```
kalpaitch: @stephenb01 these controllers are designed to be used for pages in a flow, let's think about what to do if we want a page not in a flow.

stephenb01: By the sounds of it if we want a page Not in a flow then we need to use a different controller.

kalpaitch: In https://github.com/TransformCore/beis-par-beta/pull/403 I have now moved the dashboard to just use the default base controller, which is probably what it should have done all along.

The ParBaseForm and ParBaseController are very deliberately in the par_flows module :)

stemount:
Is this PR ready? this stil has comments in it

PR #396, opened 2017-09-20
PAR-666.2: Possible other option.

kalpaitch:


stemount:
`kalpaitch wants to merge 2 commits into PAR-666 from PAR-666.2`?

PR #402, opened 2017-09-21
PAR-669: Limited the validation for errors on the cancel button.

stephenb01:


web/modules/custom/par_flows/src/Form/ParBaseForm.php
```diff
@@ -286,13 +253,61 @@ public function setIgnoredValues(array $values) {
...
   public function buildForm(array $form, FormStateInterface $form_state) {
-    $cache = array(
-      '#cache' => array(
+    // Only ever place a 'done' action by itself.
+    if ($this->getFlow()->hasAction('done')) {
+      $form['actions']['done'] = [
+        '#type' => 'submit',
+        '#name' => 'done',
+        '#value' => $this->t('Done'),
+        '#limit_validation_errors' => [],
```
stemount: is this needed for the done step?

PR #416, opened 2017-09-24
Multipass: Multi field paginated.

kalpaitch:
Well it turns out paginating plain results wasn't so complicated after all, and I think this is probably the best result we will get for now.

screenshot from 2017-09-24 01-03-58


kalpaitch:
image

stemount:
image

played @kalpaitch

PR #419, opened 2017-09-24
PAR-644: Added the invite complete page.

stephenb01:
image
 Sometimes I wish I used proper data.... :)


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsInviteForm.php
```diff
@@ -177,6 +177,15 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
       $this->getLogger($this->getLoggerChannel())->error($message, $replacements);
     }
 
+    if ($this->getFlowName() === 'partnership_application') {
```
kalpaitch: You don't need this here, all the parameters are already in the current route.

web/modules/features/par_partnership_flows/par_partnership_flows.routing.yml
```diff
@@ -500,3 +500,18 @@ par_partnership_flows.partnership_application_invite:
         type: entity:par_data_partnership
       par_data_person:
         type: entity:par_data_person
+par_partnership_flows.partnership_application_invite_sent:
+  path: '/partnership/apply/{par_data_partnership}/invite/sent/{par_data_person}'
```
kalpaitch: Also I don't think we need par_data_person in the route, we can get this as the primary business contact from the partnership.

stemount:
Changes look good, there is a usage in the branch PAR-630_blah_x which needs to be changed to work wth $blah->getUserPerson to not load an ID, rather use the returned entity. I'll change that.

PR #423, opened 2017-09-25
PAR-682: added module to limit the number of items in a multi-field.

stephenb01:


composer.json
```diff
@@ -27,6 +27,7 @@
         "drupal/csv_serialization": "~1.0.0",
         "drupal/email_registration": "~1.0.0",
         "drupal/features": "~3.5.0",
+        "drupal/field_formatter_range": "1.0",
```
kalpaitch: Would be good to use semver version constraint and tie to patch version e.g. `~1.0.0`

sync/core.entity_view_display.par_data_organisation.organisation.summary.yml
```diff
@@ -21,7 +23,11 @@ content:
     settings:
       view_mode: summary
       link: false
-    third_party_settings: {  }
+    third_party_settings:
```
kalpaitch: Will need to update the feature module now.

Also worth updating address fields on authority while you're at it too.

PR #438, opened 2017-09-27
Advice Form: The _document_ is lost after clicking 'Save' in the document type and regulatory functions form.

stemount:
screen shot 2017-09-27 at 10 16 55

Preventing the above.


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAdviceForm.php
```diff
@@ -150,7 +150,11 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
         $files_to_add[] = $file->id();
 
       }
-      $par_data_advice->set('document', $files_to_add);
+
+      // Check if any files to add from the Advice Upload form.
```
jon-edery: Check if there are any...

PR #439, opened 2017-09-27
PAR-690: Added validation for regulatory functions on advice upload.

stephenb01:


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAdviceForm.php
```diff
@@ -121,6 +121,17 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
   /**
    * {@inheritdoc}
    */
+  public function validateForm(array &$form, FormStateInterface $form_state) {
+    parent::validateForm($form, $form_state);
+
+    if (!array_keys(array_filter($form_state->getValue('regulatory_functions')))) {
```
kalpaitch: Just a point I noticed yesterday and put a @TODO somewhere in code. will need to see if we can find soe way to use entity validation to validate reference fields. It looks like it's problem with our `par_required` validation plugin. Anyway this'll do, but should put a @TODO in the code here to this effect.

PR #440, opened 2017-09-27
Accessibility: Added more unique names for action links.

kalpaitch:
- If the field is an entity reference field we can get the `label()` property of the entity itself and display `edit my entity label`.
- If the field is not an entity reference field we use the field's display name (as set in the Admin UI or BaseFieldDefinition) and display 'edit field name'.
- If there is no value set for a field we display the add button instead of the edit button, unless no add action was requested.

image

screenshot from 2017-09-27 14-44-24


web/modules/custom/par_flows/src/ParDisplayTrait.php
```diff
@@ -39,8 +39,25 @@ public function renderMarkupField($field) {
...
 
-
-  public function renderTextField($entity, $field, $view_mode, $operations = [], $single = FALSE) {
+  /**
+   * Display a FieldItemList that is not an entity reference using the given settings.
+   *
+   * Adds the required operations to each field item that gets rendered.
+   *
+   * @param string $section
+   *   The section name to display on operation links.
```
jon-edery: missing param in function definition.

PR #447, opened 2017-09-28
PAR-693: Fix for the address field error messages.

stephenb01:
This will need a lot more TLC as not all the conditions are satisfied.

image

image

image
image


web/modules/custom/par_flows/src/Form/ParBaseForm.php
```diff
@@ -398,26 +399,70 @@ public function setFieldViolations($name, FormStateInterface &$form_state, Entit
...
+              'locality' => 'town_city',
+              'postalCode' => 'postcode',
+            ];
+
+            if (!empty($element_mapping[$address_element])) {
+              $title = $form_state->getCompleteForm()[$element_mapping[$address_element]]['#title']->render();
+              $message = $this->t($violation->getMessage()
+                ->getUntranslatedString(), ['@name' => $title]);
+
+              $fragment = $this->getFormElementPageAnchor([$element_mapping[$address_element]], $form_state);
```
kalpaitch: No need to repeat this bit.

kalpaitch:
What were the issues you say you've fixed? Would be useful to assess the manner in which they're fixed, thnx.

web/modules/custom/par_flows/src/Form/ParBaseForm.php
stephenb01: Can't see how it will work without this line.

web/modules/custom/par_flows/src/Form/ParBaseForm.php
```diff
@@ -398,26 +399,70 @@ public function setFieldViolations($name, FormStateInterface &$form_state, Entit
...
+            // we use to save the item. So I have to use a special mapping
+            // array below.
+            // It appears to be getting the address field name from
+            //  /var/www/beis-par-beta/vendor/commerceguys/addressing/src/AddressFormat/AddressField.php
+            // instead of getting it from:
+            // /var/www/beis-par-beta/web/modules/contrib/address/src/Plugin/Field/FieldType/AddressItem.php
+            // There is a good chance I missed something......
+            $form_items = $form_state->getBuildInfo()['callback_object']->formItems;
+
+            $element_mapping = [
```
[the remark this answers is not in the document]
stephenb01: I agree. I just wanted to get something working so we can work from a base. But address is a composite field and I don't think we are using any other composite field.I agree. I just wanted to get something working so we can work from a base. But address is a composite field and I don't think we are using any other composite field.

stemount:
Can we make this even simpler by removing the `@name is required` totally… ?

stemount:
Would suggest to close this as not really finished and a stable release is needed today. Revisit later.

PR #455, opened 2017-09-29
feture module required to provide business logic for approving partneships

jon-edery:
This module provides all the functionality required to enable RD help desk users to approve a partnership.

### DIFF
--- a/web/modules/features/par_rd_help_desk_flows/src/Form/ParRdHelpDeskApproveForm.php
+++ b/web/modules/features/par_rd_help_desk_flows/src/Form/ParRdHelpDeskApproveForm.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace Drupal\par_rd_help_desk_flows\Form;
+
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\par_data\Entity\ParDataPartnership;
+use Drupal\par_flows\Form\ParBaseForm;
+
+/**
+ * Approving a new partnership.
+ */
+class ParRdHelpDeskApproveForm extends ParBaseForm {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected $flow = 'approve_partnership';
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFormId() {
+    return 'par_rd_help_desk_approve';
+  }
+
+  /**
+   * Helper to get all the editable values when editing or
+   * revisiting a previously edited page.
+   */
+  public function retrieveEditableValues(ParDataEnforcementNotice $par_data_enforcement_notice = NULL) {
+  }
+
+  /**
+   * {@inheritdoc}
+   */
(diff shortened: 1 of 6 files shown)


web/modules/features/par_rd_help_desk_flows/src/Form/ParRdHelpDeskApproveForm.php
```diff
@@ -0,0 +1,64 @@
...
+   */
+  public function getFormId() {
+    return 'par_rd_help_desk_approve';
+  }
+
+  /**
+   * Helper to get all the editable values when editing or
+   * revisiting a previously edited page.
+   */
+  public function retrieveEditableValues(ParDataEnforcementNotice $par_data_enforcement_notice = NULL) {
```
kalpaitch: If we could put a state value on retrieveEditableValues method, just so the values aren't stored in TempStore between different partnerships.
```
    if ($par_data_partnership) {
      // If we're editing an entity we should set the state
      // to something other than default to avoid conflicts
      // with existing versions of the same form.
      $this->setState("edit:{$par_data_partnership->id()}");
    }
```

web/modules/features/par_rd_help_desk_flows/src/Form/ParRdHelpDeskConfirmForm.php
```diff
@@ -0,0 +1,100 @@
...
+  public function getFormId() {
+    return 'par_rd_help_desk_confirm';
+  }
+
+  /**
+   * Helper to get all the editable values when editing or
+   * revisiting a previously edited page.
+   */
+  public function retrieveEditableValues() {
+
```
kalpaitch: Let's add state here too please.

web/modules/features/par_rd_help_desk_flows/src/Form/ParRdHelpDeskConfirmForm.php
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state, ParDataPartnership $par_data_partnership = NULL) {
+    $this->retrieveEditableValues();
+
+    $par_data_organisation = current($par_data_partnership->getOrganisation());
+    $par_data_authority = current($par_data_partnership->getAuthority());
+
```
kalpaitch: Probably need to check that this partnership isn't already 'active'...

web/modules/features/par_rd_help_desk_flows/src/Form/ParRdHelpDeskApproveForm.php
jon-edery: 👍

web/modules/features/par_rd_help_desk_flows/src/Form/ParRdHelpDeskConfirmForm.php
jon-edery: 👍

PR #457, opened 2017-09-30
PAR-697: Fixed flow for coordinator journey.

stephenb01:
As we are not using the *Members* section wasn't sure if I should remove the routes from the yml file and also the members section.

image


kalpaitch:
We are using the members section @stephenb01 there just aren't very many coordinated partnerships with members yet.

PR #460, opened 2017-09-30
PAR-700: Added validation to the apply check list page.

stephenb01:


stemount:
Fair play, let's get this in.

PR #461, opened 2017-09-30
remove checkbox from fields excluded from form-group-error

stemount:
Please can we double check this is needed before we get this in? Let's revisit later @nao360


stemount:
Please can we double check this is needed before we get this in? Let's revisit later @nao360

PR #483, opened 2017-10-02
PAR Entities Permissions for non-authenticated user.

stemount:
Please review/consider any security implications of this PR.

This PR fixes the /par public registry.

messages image 336386788


kalpaitch:
I'd be curious to know why legal entities/authorities and organisation showed but not regulatory functions. But I guess this is the right thing to do anyway. Let's test thoroughly.

sync/user.role.anonymous.yml
```diff
@@ -11,6 +11,14 @@ is_admin: false
 permissions:
   - 'access comments'
   - 'access content'
+  - 'access par_data_authority entities'
+  - 'access par_data_legal_entity entities'
+  - 'access par_data_organisation entities'
+  - 'access par_data_partnership entities'
+  - 'access par_data_person entities'
```
kalpaitch: I don't believe this is 100% correct. The rules about what is public are more complicated than that, it's probably only half of those you've added.

PR #485, opened 2017-10-02
Fix "when you edit a legal entity from limited company to something e…

stemount:
…lse it does not remove the company code".


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsLegalEntityForm.php
```diff
@@ -111,6 +111,12 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
 
     // Save the value for the about_partnership field.
     $legal_entity = $this->getRouteParam('par_data_legal_entity');
+
+    // Only legal entity type "limited_company" saves a companies house number.
+    if ($legal_entity && $this->getTempDataValue('legal_entity_type') !== 'limited_company') {
+      $this->setTempDataValue('company_house_no', NULL);
+    }
+
```
kalpaitch: Just conditionally set the value as empty in the else statement on line 137.

PR #486, opened 2017-10-03
Performance: Removing all sqlsrv and migration modules.

kalpaitch:
In order to run the webprofiler I needed to remove all mention of the sqlsrv module.


stemount:
merged :medal:

PR #490, opened 2017-10-03
PAR-745: Only organisations can edit legal entities on the apply jour…

stephenb01:
…ney.


sync/user.role.par_helpdesk.yml
```diff
@@ -83,7 +81,6 @@ permissions:
   - 'edit par_data_authority entities'
   - 'edit par_data_enforcement_notice entities'
   - 'edit par_data_inspection_plan entities'
-  - 'edit par_data_legal_entity entities'
```
jon-edery: Are we sure we want to remove edit legal entities from the par help desk role?

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsDetailsForm.php
```diff
@@ -85,18 +86,34 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
...
+    if ($checkbox === 'partnership_info_agreed_business' && !$par_data_partnership->getBoolean($checkbox)) {
+      // Get current user and check permissions.
+      $account = $this->getCurrentUser();
+      if ($account->hasPermission('edit par_data_legal_entity entities')) {
+        $operations[] = 'edit-entity';
+      }
+      if ($account->hasPermission('add par_data_legal_entity entities')) {
+        $operations[] = 'add';
+      }
+    }
```
jon-edery: Are we missing  the following: 
 if ($account->hasPermission('delete par_data_legal_entity entities)) {
       $operations[] = 'delete';
      }

sync/user.role.par_helpdesk.yml
stephenb01: Yes. The is the message I got from @nao360. The ONLY time they can add/edit legal entities is during the apply stage. After that it is set in stone.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsDetailsForm.php
stephenb01: Yes we should have a delete here, but not sure if delete has been implemented yet or if this is in another jira ticket.

kalpaitch:
@nao360 We need to discuss this requirement. It was requested very urgently yesterday but needs to be thought of much more carefully. It goes against many of the decisions we'd made up to this point, primarily that the business confirming a partnership and a business updating a partnership were the same journey.

PR #493, opened 2017-10-04
PAR-558: Added cache warmer to production.

kalpaitch:
This adds a cache warmer to all deployments, which reduces the time taken for `clare.scawthorn@london-fire.gov.uk` to log in to a non-cached site from above 60 seconds to 10-15 seconds.

This is an improvement, but we still need tickets to address the real causes of bad performance and profile some of the weaker points.

At present it seems one of the big weaknesses of the current system is the number of duplicated and redundant data items inherited from PAR 2 so we may need data clean up/de-duping tickets also.


chris-moreton:
Looks good. I'll soon need to go through a refactoring phase to organise the various shell scripts and other devops-related tools in the repo, but for now, those scripts will just sit in the root directory.

PR #494, opened 2017-10-05
Setup reverse proxy settings in settings.php

chris-moreton:
The solution here was to add 127.0.0.1 also to the list of trusted proxies. As localhost is the final "proxy" to forward the request, Symfony was not going any further up the chain. When 127.0.0.1 is specified along with all the AWS Cloud Front proxy addresses, the Symfony component will walk up the chain until it finds an IP that it doesn't recognise as a proxy, which should be the client IP.


kalpaitch:
Cool, I believe you :)

PR #495, opened 2017-10-05
Update partnership report with correct script and SQL

chris-moreton:
1) As were were using the outer join, not having some advice or inspection plan or legal entity is not a problem. However, there is a problem if you have some of those things and all of them are in "deleted" or "revoked" status - in this case, rather than bringing back an empty row, it simply ignores the partnership, just as if you had done an inner join when there were no linked rows in the child tables. To solve, we don't filter on the "revoked" or "deleted" status, but instead bring back that status so that the PHP reporting script can do what needs to be done in each case.
2) Removed a daft "else" clause that meant the first legal entity and advice for each partnership never got added to the report.
3) I really want to do this properly. I wonder if I can do it in this sprint.


chris-moreton:
@kalpaitch I agree, although it's probably always safest to review anything that gets checked into master, just in case something daft gets added.

PR #497, opened 2017-10-09
PAR-742 PAR-743 Only show active partnerships in Search Partnerships, PAR register.

stemount:
Can be applied to any field that has allowed options/values.

Provides allowed options in the view filter form.

messages image 3196204343

messages image 960914596


web/modules/custom/par_data/src/Plugin/views/filter/ParFieldAllowedOptionsFilter.php
```diff
@@ -0,0 +1,63 @@
...
+   * {@inheritdoc)
+   */
+  public function query() {
+    // This filter does not apply if not on a PAR entity.
+    if (!$this->par_data_manager->getParEntityType($this->getEntityType())) {
+      return;
+    }
+
+    // Get data table.
+    $data_table = $this->par_data_manager->getParEntityType($this->getEntityType())->get('data_table');
```
kalpaitch: If we could determine the table based on whether the entity is being gathered from a relationship, similar to the way we did it for the membership filter... it would allow us to filter based on the status of an enforcement action...

stemount: Good suggestion, Gonna give it a whirl

web/modules/custom/par_data/src/Views/ParDataViewsData.php
```diff
@@ -47,7 +47,16 @@ public function getViewsData() {
       ],
     );
 
-    // Add the current company computed field to Views.
+    // PAR Partnership Status Filter.
```
kalpaitch: Let's not specifically put this on partnerships, we need this for enforcement notices and actions also.

web/modules/custom/par_data/src/Views/ParDataViewsData.php
```diff
@@ -47,7 +47,16 @@ public function getViewsData() {
...
+        'title' => t('PAR Partnership Status'),
+        'help' => t('Provides a partnership status field filter for PAR partnerships.'),
+        'id' => 'par_data_field_allowed_options_filter',
+      ],
+    ];
+
+    // PAR Status Field and Filter.
     $data[$this->entityType->getDataTable()]['par_status'] = [
       'title' => t('PAR Status'),
       'field' => [
```
kalpaitch: You can put the 'filter' as a separate key here I think.

web/modules/custom/par_data/src/Plugin/views/filter/ParFieldAllowedOptionsFilter.php
kalpaitch:
```
// The normal use of ensureMyTable() here breaks Views.
    // So instead we trick the filter into using the alias of the base table.
    // @see https://www.drupal.org/node/271833.
    // If a relationship is set, we must use the alias it provides.
    if (!empty($this->relationship)) {
      $this->tableAlias = $this->relationship;
    }
    // If no relationship, then use the alias of the base table.
    else {
      $this->tableAlias = $this->query->ensureTable($this->view->storage->get('base_table'));
    }
    // Get field to query e.g. "par_partnerships_field_data.id".
    $id = $par_entity_type->getKeys()['id'];
    $revision_table = "{$this->tableAlias}.{$id}";
    // Where filter on partnership id to those the user is allowed to update.
    $this->query->addWhere(0, $revision_table, $this->value, 'in');
```

stemount:
@kalpaitch there are no specific targets for filtering for partnership, it will now be on all relevant entities, should this be of any use.

I also put in the enforcement action exposed status filter, have confirmed it is getting the correct data from config files, but these may need to change.

messages image 2822673926

web/modules/custom/par_data/src/Views/ParDataViewsData.php
```diff
@@ -37,17 +37,25 @@ public function getViewsData() {
...
+      'filter' => [
+        'title' => t('PAR Partnership Status'),
+        'help' => t('Provides a partnership status field filter with options.'),
+        'id' => 'par_data_field_allowed_options_filter',
       ],
-    );
+    ];
+
+    // PAR Enforcement Action Status Filter.
+    $data[$this->entityType->getDataTable()]['enforcement_action_status'] = [
```
kalpaitch: this will put an 'enforcement_action-status' on every entity type...

kalpaitch: Would this not work??
```
$data[$this->entityType->getDataTable()]['par_status'] = [
  .....
  'filter' => [
    'title' => t('PAR Status'),
    'help' => t('Provides a enforcement status field filter with options.'),
    'id' => 'par_data_field_allowed_options_filter',
  ],
];
```

kalpaitch:
Looks awesome @stemount, like the filter..

stemount:
@kalpaitch sorry for to/from… back to what it is.

web/modules/custom/par_data/src/ParDataManager.php
```diff
@@ -131,7 +131,7 @@ public function getParBundleEntity(string $type, $bundle = NULL) {
     $entity_type = $this->getParEntityType($type);
     $definition = $entity_type ? $this->getEntityBundleDefinition($entity_type) : NULL;
     $bundles = $definition ? $this->getEntityTypeStorage($definition->id())->loadMultiple() : [];
-    return $bundles && isset($bundles[$bundle]) ? $bundles[$bundle] : current($bundles);
+    return $bundles && isset($bundles[$bundle]) && !empty($bundles[$bundle]) ? $bundles[$bundle] : current($bundles);
```
jon-edery: Above this statement you can call id() on NULL if the if statement for $definition returns null and the $entity_type is set to NULL.  you could check the passed in value for $type within this function is within the allowed system entities list to prevent an object function being called on NULL system error.

PR #500, opened 2017-10-11
PAR-795: Create new premises and add to partnership.

stephenb01:
There are a few areas I'm not sure if they are right.
The first is the search criteria for the address. This is based on the first line of the address and the postcode.

The other issue is the use of `$this->retrieveEditableValues($par_data_partnership);`. Shouldn't need to use this but if this is not here then `$this->getDefaultValues('address_line1', '', 'par_partnership_address')` will return an empty value.


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAddressForm.php
```diff
@@ -199,6 +199,41 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
           ->error($message, $replacements);
       }
     }
+    else {
```
stemount: How about switching out premises to minimise the lines of code?

`$premises = $this->getRouteParam('par_data_premises')  ?: ParDataPremises::create(['type' => 'blah'])` or something even better than this - there are other examples of this in the codebase

stemount:
Trying hard to understand the PR 😀 e.g. `There are a few areas I'm not sure if they are right.` :)

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAddressForm.php
kalpaitch: Yes please...

web/modules/custom/par_flows/src/ParDisplayTrait.php
```diff
@@ -247,7 +248,7 @@ public function renderSection($section, $entity, $fields, $operations = [], $tit
 
       // Only add the add link if it is in the allowed operations.
       $link_name_suffix = strtolower($field->getFieldDefinition()->getLabel());
-      if (isset($operations) && (in_array('add', $operations))) {
+      if (isset($operations) && (in_array('add', $operations)) && !($single && !$field->isEmpty())) {
```
stemount: is $single defined?

stephenb01: $single is a parameter passed to the function:
*   public function renderSection($section, $entity, $fields, $operations = [], $title = TRUE, $single = FALSE) { *

stemount: I don't know if we need this $single_item change… because 'add' will not be in $operations if it is a single item (well, it shouldn't be!)

stemount:
I fiddled with this PR a little bit to prevent any more to/from on a PR that is a week old (the value is strained everytime!)

stephenb01:
@stemount By changing $single_item to $single you will not be able to add more Trading Names.
This is using $single:
image

This is using $single_item:
image

This also applies to the SIC codes, although I'm sure you are only allowed one, but @nao360 can confirm this.

Again contacts is also affected by this change. This is using $single:
image

This is using $single_item:
image

$single is changed not because there is only meant to be one value but because there is only one value at present. So there could be one contact but should we allow the user to add more? Trading names I know we should and if this is different from contacts and SIC codes then this probably needs to be addressed differently.
`        if ($field->count() <= 1) {
          $single = TRUE;
        }`

Hope the above is clear!!

stemount:
@stephenb01 Damn. Doesn't pass the tests. I reverted my messing around.

stemount:
my findings are that the Organisation does not save properly if you are entering a new Organisation.

messages image 3660198503

it cannot find a organisation on the partnership suggesting the previous step either does not create (or does not link) the organisation. 

messages image 3395001765

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAddressForm.php
```diff
@@ -187,6 +188,15 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
       $premises->set('nation', $this->getTempDataValue('country'));
 
       if ($premises->save()) {
+        // Add premises to partnership if a new PAR Premises record is created.
+        if (!$this->getRouteParam('par_data_premises')) {
+          $par_data_partnership = $this->getRouteParam('par_data_partnership');
```
stemount: Does this work for an application journey?

stemount:
*existing record shows an edit link*

messages image 3660198503

*added a Organisation address header to be more uniform*
messages image 33436972

Please review against the AC of the ticket, it doesn't meet all the A/C as it stands.

PR #504, opened 2017-10-12
PAR-620: Fix for displaying number of employees band.

stephenb01:
Ideally the field value should be the ID of allowed values and we convert that into the actual value the user sees. However in this case, and for the number of members band value, we are actually storing the value itself and not the ID. So may need a separate task to identify why we are saving the value and not the ID, convert the existing ones to use ID and then display it correctly.

So if we have 250+ displayed for number of employees we should be storing it as the ID for label 250+ which is 1000. We are storing it as 250+ in the field itself.
 
image


kalpaitch:
Total apologies @stephenb01 I see what you mean... so this is entirely a data issue.

* All the data that was migrated was the value stored
* All the data that has been updated or added since then has stored the key (ID)

image

I think if we put this patch in the latter will not display correctly and you're totally right we do need to fix the data.

stemount:
indeed. let's keep this as the list formatter. we could actually make the list formatter display a string if absolutely necessary.

kalpaitch:
Yea we could even put allow the formatter to be configurable, so that we could tell it what to do if the value is not allowed. This would allow us not to correct the data for now...

stemount:
yeah, lets make this configurable as per https://www.drupal.org/docs/8/creating-custom-modules/create-a-custom-field-formatter. have a configurable field that is either "display string if no match" or "display custom text"

PR #514, opened 2017-10-16
PAR-620: Fix for displaying number of members band.

stephenb01:
Added a configuration screen to the Par data formatter. It will allow the user to decide how to manage the values that cannot be found in the list of allowed values.

image

image


web/modules/custom/par_data/src/Entity/ParDataType.php
```diff
@@ -179,7 +179,16 @@ public function getAllowedValues($field_name) {
    */
   public function getAllowedFieldlabel($field_name, $value = FALSE) {
     $allowed_values = $this->getConfigurationElementByType($field_name, 'allowed_values');
-    return isset($allowed_values[$value]) ? $allowed_values[$value] : FALSE;
+
+    if (isset($allowed_values[$value])) {
+      return $allowed_values[$value];
+    }
+    // Need to check if the value is an actual value.
```
kalpaitch: No please, we shouldn't catch this situation, it's not wanted, not here.

web/modules/custom/par_data/src/Plugin/Field/FieldFormatter/ParListFormatter.php
```diff
@@ -65,13 +66,55 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
+      '#title' => t('Default option when there is no match'),
+      '#type' => 'select',
+      '#options' => [
+        'string' => $this->t('Display string value if no match'),
+        'custom' => $this->t('Display custom text'),
+      ],
+      '#default_value' => $this->getSetting('default_option'),
+    ];
+
+    $element['custom_text'] = [
```
kalpaitch: I'm not sure what the use case for 'custom_text' is because it's not being used. Seems like it defeats the point of a list??

stemount:
Custom text - This may have been my doing just guessing random requirements. We could just have a "show the value even if it doesn't match" checkbox, it might be better.

stemount:
just a query about a removal of unknwn value

web/modules/custom/par_data/src/Plugin/Field/FieldFormatter/ParListFormatter.php
```diff
@@ -65,13 +66,39 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
       $value = $bundle_entity->getAllowedFieldlabel($field_name, $item->value);
       // Render each element as markup.
+      if (!$value && $this->getSetting('display_original_value')) {
+        $value = $item->value;
+      }
+
       $element[$delta] = [
         '#type' => 'markup',
-        '#markup' => $value ? $value : 'Unknown value',
+        '#markup' => $value,
```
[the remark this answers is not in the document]
stephenb01: Well spotted!!!

web/modules/custom/par_data/src/Plugin/Field/FieldFormatter/ParListFormatter.php
```diff
@@ -74,4 +79,26 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsForm(array $form, FormStateInterface $form_state) {
+    $element['display_original_value'] = [
+      '#title' => t('Display original value if there is no match'),
+      '#type' => 'checkbox',
+      '#default_value' => $this->getSetting('default_option'),
```
stemount: should this be `display_original_value`?

stemount:
This looks cool to me @stephenb01 just one Q and then if not this is approved

stephenb01:
Well spotted! Should be fixed now....

PR #519, opened 2017-10-19
Mem gate

chris-moreton:
Modify development settings to match the memory settings being used on Travis and being used temporarily in the php.ini files on PaaS. Once the relationship-building routine has been rewritten/refactored to use less memory, this setting will again be removed.


stemount:
There are other memory intensive operations, like configuration import and enabling test data can run out of RAM. it's not just for the partnership search @chris-moreton

PR #525, opened 2017-10-23
wonder if these run first time

stemount:
i propose these tweaks as it makes the build time quicker.


stemount:
this is in the other PR now. closing this PR.

PR #526, opened 2017-10-24
PAR-865 added correct config mapping for approved enforcement actions

jon-edery:
This config update fixes the approve state reffrences for enforcement actions as in the backend we are saving "approved" to indicate an approved action not "allowed".

I have kept the old reference in case this is needed for data that has been migrated as allowed to ensure this does not cause any regressions.


kalpaitch:
Think better to correct the mistake and have only one state for 'Approved', traditionally this was `allowed`. It would defeat the point of having knowing what state an action is in having this. Can do hook update to correct any incorrectly entered data.

jon-edery:
Agreed ill get the latest version of the live db today check the data update if required and remove the old reference.

stemount:
yea, we definitely need a new database dump.

jon-edery:
Fix added in PAR-421 branch.

